PR #21, opened 2026-04-14
feat: stream MKV playback without full remux wait

akvasha:
## Summary

Opening an MKV in the built-in player previously blocked for 10–20 seconds behind a "Preparing MKV for playback…" overlay while ffmpeg remuxed the entire file to a temp MP4 via `-c copy -movflags +faststart`.

This PR replaces that with a streaming MSE (Media Source Extensions) pipeline: ffmpeg produces a fragmented MP4 to stdout, the main process forwards chunks over IPC, and the renderer feeds them into a `MediaSource` `SourceBuffer`. Playback starts in ~1 s. Unbuffered seeks are serviced by killing and respawning ffmpeg with `-ss`.

### Main process

- `player:remux-mkv-stream(mkvPath, initialSeek?)` — spawns ffmpeg with fragmented MP4 output (`+frag_keyframe+empty_moov+default_base_moof+separate_moof`, `-frag_duration 1000000`) and timestamp normalization (`-fflags +genpts`, `-avoid_negative_ts make_zero`, `-muxpreload 0`, `-muxdelay 0`). Resolves once the MP4 header prelude is on the wire; returns `{ sessionId, duration, mimeType, hasSubtitlesPending, initialSeek }`.
- Session state (`MseSession`) holds the prelude buffer, pending chunks, generation counter, and child process handle.
- `player:stream-start(sessionId)` — flushes the prelude plus any queued chunks after the renderer has attached `MediaSource`.
- `player:stream-ack(sessionId, bytes)` — backpressure signal. Main pauses `stdout` above a high-watermark and resumes below a low-watermark to prevent unbounded memory growth.
- `player:stream-seek(sessionId, seekSeconds)` — bumps `session.generation`, SIGKILLs the current ffmpeg, clears prelude and queued chunks, respawns with `-ss seekSeconds`. Generation check in every stdout/stderr/exit handler drops stale chunks from the killed process.
- Subtitle extraction (ffprobe + `-map 0:s:0 -c:s ass`) runs in parallel and is pushed to the renderer via `player:stream-subtitles` when ready.
- `player:cleanup-remux` kills any active session and clears temp state.

### Renderer (`PlayerView.vue`)

- MSE wiring: `startMseSession` creates a `MediaSource`, attaches it via `URL.createObjectURL`, adds a `SourceBuffer` for the negotiated `mimeType`, sets `sb.timestampOffset = initialSeek` before the first chunk, and calls `playerStreamStart` to flush the prelude.
- `onPlayerStreamChunk` appends to a queue; `pumpAppendQueue` drains it with a `MAX_BUFFER_AHEAD` of 30 s (prevents `QuotaExceededError` piling up too far past the playhead).
- On quota hit: evict only ranges behind the playhead (keeps the nearest keyframe for current playback).
- Unbuffered seek: 250 ms-debounced on the `seeking` event. Calls `sb.abort()` (resets parser state), removes all buffered ranges, sets `timestampOffset = newPosition`, then calls `playerStreamSeek` → `playerStreamStart`. The `unbufferedSeekInFlight` flag blocks further appends until the new prelude arrives, avoiding audio overlap from the old segment.
- Rapid arrow-key handling: the debounce collapses spam into one respawn; the seeking handler short-circuits when the target is already inside `sb.buffered` or when `sb.buffered.length === 0` (just respawned — don't respawn again).
- Resume-from-timestamp: reads `watchProgressGet` in `prepareMkvForPlayback` and passes `saved.position - 1` as `initialSeek` so ffmpeg starts at the saved position directly. `resumeFromSavedPosition` skips its seek for MSE streams since the video is already at the right position.
- Template: `:src="videoSrc || undefined"` avoids an empty string setting `HTMLMediaElement.error` to `MEDIA_ERR_SRC_NOT_SUPPORTED` during tear-down.
- Fallback: if `playerRemuxMkvStream` errors, the renderer falls back to the legacy `playerRemuxMkv` (full remux) path and keeps playing.

## IPC surface

New channels:
- `player:remux-mkv-stream(mkvPath, initialSeek?)` → session info
- `player:stream-start(sessionId)`
- `player:stream-ack(sessionId, bytes)`
- `player:stream-seek(sessionId, seekSeconds)` → `{ ok: true } | { error }`

New events (main → renderer):
- `player:stream-chunk` — `{ sessionId, data: Uint8Array }`
- `player:stream-end` — `{ sessionId }`
- `player:stream-error` — `{ sessionId, error }`
- `player:stream-subtitles` — `{ sessionId, content }`

Updated: `player:cleanup-remux` now also terminates active streaming sessions.

## Files changed

| File | What changed |
|------|--------------|
| `src/main/index.ts` | Streaming session state, 5 new IPC handlers, ffmpeg respawn with generation counter, subtitle extraction wired to event |
| `src/preload/index.ts`, `src/preload/types.d.ts` | Bridges + types for the new IPC surface |
| `src/renderer/src/components/PlayerView.vue` | MSE wiring, append queue, quota eviction, debounced unbuffered seek, MKV fallback |
| `DESIGN.md` | Architecture notes for the streaming path |
| `TODO.md` | Strike the done item |
| `package.json` | Version bump 3.5.0 → 3.6.0 |

## Rollback

Legacy `player:remux-mkv` handler is preserved. If the MSE path breaks, remove the `playerRemuxMkvStream` call in `prepareMkvForPlayback` and only call `playerRemuxMkv`.

## Test plan

- [x] MKV opens in <2 s, no blocking overlay
- [x] Forward/backward seek inside buffered region (instant)
- [x] Forward seek past buffered region (ffmpeg respawns)
- [x] Backward seek past buffered region
- [x] Rapid arrow-key hold (no freeze, no respawn storm)
- [x] Resume from saved watch-progress timestamp
- [x] Translation / next-episode switch mid-playback (old session torn down, new session starts)
- [x] Subtitles appear shortly after playback starts
- [x] MP4 playback unaffected
- [x] Fallback path still works if streaming init fails

Generated with Claude Code

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -295,8 +295,14 @@ LibraryView shows both with indicators:
 | `player:get-stream-url` | invoke | Fetch CDN stream URL + raw ASS subtitle content + all available stream qualities for a translation |
 | `player:get-local-subtitles` | invoke | Read local .ass file alongside video, return raw ASS content |
 | `player:find-local-file` | invoke | Find local file path for a translation by animeName/episodeInt/translationId |
-| `player:remux-mkv` | invoke | Remux MKV→MP4 via ffmpeg stream copy to temp dir for HTML5 playback |
-| `player:cleanup-remux` | invoke | Delete temp remuxed files when player closes |
+| `player:remux-mkv` | invoke | Legacy: remux MKV→MP4 via ffmpeg stream copy, await completion before returning (used as fallback when codecs aren't MSE-compatible) |
+| `player:remux-mkv-stream` | invoke | ffprobe MKV + start fragmented-MP4 pipe from ffmpeg; returns `{ sessionId, duration, mimeType, hasSubtitlesPending }` for MSE consumption |
+| `player:stream-chunk` | event (main→renderer) | Fragmented-MP4 bytes from an active session, appended into the renderer's `SourceBuffer` |
+| `player:stream-end` | event (main→renderer) | ffmpeg finished writing; renderer calls `mediaSource.endOfStream()` after draining its queue |
+| `player:stream-error` | event (main→renderer) | ffmpeg exited non-zero or spawn failed; renderer falls back to legacy remux |
+| `player:stream-ack` | invoke | Renderer reports bytes it has consumed from the session; main resumes ffmpeg stdout once pending bytes drop below the low watermark |
+| `player:stream-subtitles` | event (main→renderer) | Subtitle `.ass` content extracted from an MKV stream session, pushed when extraction finishes |
+| `player:cleanup-remux` | invoke | Kill any active stream sessions and delete all temp remuxed files |
 | `shell:open-external` | invoke | Open URL in default browser (returns success boolean) |
 
 ## Key Types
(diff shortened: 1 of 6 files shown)


.github/workflows/check.yml
mikkerlo, line 1: Why this file is removed? Return it to the state before the commit, it shouldn't affect PR

.github/workflows/claude.yml
mikkerlo, line 1: Why this file is removed? Return it to the state before the commit, it shouldn't affect PR

mikkerlo (approved):

PR #27, opened 2026-04-15
feat: HEVC (H.265) support in MSE streaming path (v3.7.0)

akvasha:
## Summary
- Add `hevcCodecString` producing `hvc1.<profile>.<compat>.L<level>.B0` for HEVC Main / Main 10 / Main Still Picture.
- Extend `probeMkvForMse` to try AVC first, then HEVC; return the chosen codec so the ffmpeg spawn can adapt.
- Emit `-tag:v hvc1` on the ffmpeg command when the input video is HEVC — Chromium MSE rejects the default `hev1` tag (parameter sets inline) and requires `hvc1` (parameter sets in the sample entry).
- Log the negotiated MIME + `isTypeSupported` result in the renderer for easier diagnosis.
- Bump version to 3.7.0 and update DESIGN.md / TODO.md.

## Context
Before this change, HEVC MKVs always fell back to the legacy full-remux path (10–20 s blocking, and unplayable on Linux Electron). Many anime releases ship as HEVC, so the MSE path was missing a substantial subset of the library.

The fallback behaviour is unchanged: when `MediaSource.isTypeSupported('video/mp4; codecs="hvc1.…, mp4a.…")` returns false (e.g., a Linux build with HEVC disabled), the renderer still drops to `player:remux-mkv`.

## Test plan
- [ ] H.264 MKV — plays via MSE (regression check on the v3.6.0 path).
- [ ] HEVC Main 10 MKV on Windows — plays via MSE, no "Preparing MKV…" overlay.
- [ ] HEVC MKV on Linux — renderer logs `MSE ... supported=false` and falls back to legacy remux.
- [ ] Rapid seek on an HEVC file — no duration drift / stale-chunk bug (generation filter covers this).


akvasha:
## HEVC playback investigation — platform decoder limitation

After testing this branch end-to-end with an HEVC MKV, the code path works correctly but playback fails on Linux/WSL2 because Chromium has no HEVC decoder available. Audio plays, video is a black screen — classic "container parsed, no decoder" signature.

### Diagnostic

Added a temporary probe at `ready-to-show` that runs `MediaSource.isTypeSupported()` against several HEVC codec strings. Output on this machine (Electron 41, Linux x64 under WSL2):

```
[mse-probe] YES video/mp4; codecs="avc1.640028, mp4a.40.2"
[mse-probe] NO  video/mp4; codecs="hvc1.1.6.L120.B0, mp4a.40.2"
[mse-probe] NO  video/mp4; codecs="hvc1.2.4.L120.B0, mp4a.40.2"
[mse-probe] NO  video/mp4; codecs="hev1.1.6.L120.B0, mp4a.40.2"
```

Also tried `app.commandLine.appendSwitch('enable-features', 'PlatformHEVCDecoderSupport')` — no change. Chromium's HEVC support is gated on a platform decoder (Media Foundation on Windows, VideoToolbox on macOS, system IMF/V4L2 on Linux with hardware); stock Linux Chromium builds ship no software HEVC decoder at all, so `isTypeSupported` returns false regardless of flags.

### What this means for the PR

- The `hvc1.…` codec string generation and `-tag:v hvc1` mux flag are correct per ISO/IEC 14496-15. When MSE *does* accept the mime, the fragmented MP4 it produces plays.
- Expected platforms: Windows (Electron's bundled Chromium links against MF HEVC when HW supports it) and macOS (VideoToolbox). Can someone with a Windows or macOS build confirm playback of an HEVC Main or Main 10 source?
- Linux users will continue to hit the fallback. Currently the fallback is the legacy full-remux path, which also writes HEVC into an MP4 — same decoder problem, just delayed.

### Options for the Linux gap

1. **Accept the limitation.** Document that HEVC playback requires Windows/macOS. Simplest; matches upstream Chromium posture.
2. **Transcode-to-H.264 fallback.** When `MediaSource.isTypeSupported(hvcMime) === false`, respawn ffmpeg with `-c:v libx264 -preset ultrafast -tune zerolatency -c:a copy` into the same fragmented-MP4 pipe. CPU-heavy but universal; reuses all the existing MSE / seek / generation-counter plumbing.
3. **External player on Linux.** Launch `mpv` / `vlc` with the local file when HEVC is detected and MSE rejects it. Cheap but loses shader/subtitle integration.

My suggestion: land the PR with option (1) (add a user-facing "HEVC requires Windows/macOS on this build" notice when `isTypeSupported` fails and we fall back) and file option (2) as a follow-up TODO. Option (3) is a last resort.

The `[mse-probe]` diagnostic and `PlatformHEVCDecoderSupport` flag from this investigation are not committed — happy to push them behind a dev-only toggle if useful for others to reproduce.

mikkerlo (commented):
Great implementation of HEVC support in the MSE path! The  helper and the  flag are spot on for compatibility.

A couple of suggestions for improvement:
1. **Warning for Unsupported Platforms**: On platforms where HEVC is not supported (e.g., Linux without the feature flag), we should probably show a one-time warning to the user when they first attempt to play an HEVC file that requires a full transcode/remux, as this will be slow.
2. **Settings Warning**: It would be helpful to add a warning message in  when an HEVC codec is selected for merging, informing the user about potential playback compatibility issues on some platforms.

mikkerlo (commented):
Great implementation of HEVC support in the MSE path! The `hevcCodecString` helper and the `-tag:v hvc1` flag are spot on for compatibility.

A couple of suggestions for improvement:
1. **Warning for Unsupported Platforms**: On platforms where HEVC is not supported (e.g., Linux without the feature flag), we should probably show a one-time warning to the user when they first attempt to play an HEVC file that requires a full transcode/remux, as this will be slow.
2. **Settings Warning**: It would be helpful to add a warning message in `SettingsView.vue` when an HEVC codec is selected for merging, informing the user about potential playback compatibility issues on some platforms.

mikkerlo (commented):
Following up on the previous comment, here are some specific implementation suggestions:

### 1. User-visible warning in PlayerView.vue
When MSE fails and we fall back to the legacy remux path at line 343 of `src/renderer/src/components/PlayerView.vue`, it would be great to show a small notice if it's an HEVC file. Something like: 'HEVC is not natively supported on this platform. Remuxing might be slow and may not result in playable video.'

### 2. Transcoding in player:remux-mkv
Currently, `player:remux-mkv` uses `-c copy`. If the platform doesn't support HEVC, the resulting MP4 will still be unplayable. We might want to consider allowing a full transcode (e.g., using `-c:v libx264`) as a last resort, though that would be much slower.

### 3. Warning in SettingsView.vue
Even though `SettingsView.vue` isn't in this PR, we should add a warning label when a user selects an HEVC encoder (H.265) in the 'Merging' tab. This will help users understand that while it saves space, it might cause playback issues on some Linux systems or older hardware.

mikkerlo:
В настройках в авто мердже, есть опция перекодировать в h265, если эту опцию выбирают то давай показывать баннер что будут проблемы с воспроизведением, и спрашивать правда ли он их хотят.

akvasha:
Thanks @mikkerlo! Addressed (1) and (3) in `67804a5` — deferred (2) to follow-up.

### (1) PlayerView fallback notice — implemented
When MSE rejects the HEVC mime (`hvc1` / `hev1`) and we drop to the legacy full-remux path, the "Preparing MKV for playback" overlay now shows an amber line:

> HEVC (H.265) is not natively supported on this platform — the remuxed file may fail to play.

This only fires when the mime came back HEVC-shaped, so H.264 fallbacks (e.g. some exotic audio codec) don't get the warning.

### (3) SettingsView codec warning — implemented
When `libx265` or any `hevc_*` hardware encoder is selected in the Merging tab, an amber hint appears under the dropdown:

> H.265 merges save disk space, but playback requires a platform HEVC decoder. Linux builds without one will need to re-encode on playback or use an external player.

### (2) Transcode-to-H.264 fallback — deferred
Agree this is the real fix for Linux, but landing it here would bloat the PR: it needs real-time re-encode in both the MSE pipe and the legacy remux, a consent prompt (libx264 ultrafast still burns CPU + battery on laptops), progress surfacing, and probably a settings toggle for "always transcode HEVC". Better as its own PR once this one lands. Happy to file the TODO entry in a separate commit if you'd like it tracked before merge.

mikkerlo:
Если на линуксе будет срабатывать старый механизм тогда, и видео будет отображаться, меня это устроит. Добавь тогда это в todo, используя команду /add-todo

akvasha:
Спасибо за апдейт! Поправил:

**Confirmation banner только когда это релевантно** (`8ddcbf9`)
- `SettingsView.vue` теперь при загрузке проверяет `MediaSource.isTypeSupported('video/mp4; codecs="hvc1.…"')` один раз. На Windows/macOS с рабочим HEVC-декодером — ни inline-предупреждения, ни диалога: пользователи выбирают H.265 без шума.
- На Linux без декодера: при смене кодека на любой `libx265` / `hevc_*` показывается `window.confirm` с объяснением, что файл будет выдавать чёрный экран + звук в встроенном плеере и нужен внешний. Если отказываются — значение откатывается к предыдущему.
- Inline amber-hint под dropdown тоже гейтится тем же флагом.

**Небольшое замечание по второй части твоего комментария:** легаси-механизм (`player:remux-mkv` → `-c copy`) на Linux для HEVC не спасает — он меняет контейнер, а не кодек. Результирующий MP4 всё так же не декодится `<video>`. Т.е. пока там действительно видно только чёрный экран. Настоящий фикс — real-time transcode в H.264.

**TODO добавил** (`c216d93`, `TODO.md` → `## 1. HEVC → H.264 transcode fallback…`). План включает:
- где гейтить (в `prepareMkvForPlayback`, когда `isTypeSupported === false` И mime содержит `hvc1`/`hev1`);
- новый IPC `player:remux-mkv-stream-transcode` через 4-файловый паттерн, переиспользующий `MseSession` + generation-counter (они кодек-агностичны);
- стратегию аудио (copy для AAC, транскод в AAC для AC-3/E-AC-3/DTS/FLAC);
- приоритет h264-энкодеров (`h264_vaapi` → `h264_nvenc` → `h264_qsv` → `libx264`) с обязательным 1-frame dry-run пробингом, потому что наличие в `-encoders` не означает что энкодер работает (например `h264_vaapi` без `/dev/dri/renderD128`);
- новую настройку `hevcTranscodeOnPlay: 'ask' | 'always' | 'never'` + модалку с выбором "Transcode / Always / Open external";
- отдельный Linux-блок caveats (почему `PlatformHEVCDecoderSupport` на WSL2 бесполезен; как правильно пробить `h264_vaapi`).

akvasha:
Reviewed the three Copilot findings.

**1. `hevcPlaybackSupported` should use `canPlayType`, not `isTypeSupported` — accepted, fixed in `665fcceb`.**
Legitimate point: merged H.265 files play through a plain `<video src="…">` element (not MSE), and the two support surfaces can diverge in Chromium. Switched the probe to `HTMLVideoElement.canPlayType` and included both `hvc1` and `hev1` tags for Main / Main 10. Warning now tracks what the player will actually attempt.

**2. HEVC `stream.level` comment is self-contradictory — already fixed** by Copilot Autofix in `d02785bc`. Comment now correctly describes ffprobe's HEVC `level` field as the raw `level_idc` (e.g. 120 → 4.0) and why we emit it verbatim after `L`.

**3. DESIGN.md temp dir note is incomplete — already fixed** by Copilot Autofix in `e8f4b76`. The line now says the temp dir holds both remuxed `.mp4` files and extracted `.ass` subtitles.

mikkerlo (approved):

PR #29, opened 2026-04-15
Player UX: volume persistence, streaming toast, subtitle auto-pick

akvasha:
## Summary
- **Volume persistence**: player volume + mute are now saved under new `playerVolume` / `playerMuted` settings (400 ms debounced writes) and restored on every mount, so adjusting the slider once sticks across episodes and restarts.
- **Streaming toast**: the *Streaming from server* banner auto-dismisses after 3.5 s instead of sitting on screen for the whole episode.
- **Subtitle/translation auto-pick**: when opening an anime with no per-anime translation preference stored, the detail page counts `(translationType, author)` pairs across `downloadedEpisodes` metadata and defaults the selection to the most common pair — so if you have *English Subtitles — Anime Time* downloaded, the page opens on that instead of the global `subRu` default.

## Test plan
- [x] `npm run typecheck`
- [x] Volume: set slider mid-range on episode 1, switch to episode 2 → slider keeps the value. Restart app → slider still at that value.
- [x] Streaming: open a CDN-streamed episode → banner appears, fades out after ~3.5 s, does not reappear on play/pause.
- [x] Auto-pick: on an anime with English Anime Time downloads but no stored prefs, page opens with `subEn` + Anime Time selected (previously was `subRu` + SovetRomantica).
- [x] No regression on anime with an explicit stored translation pref — those still honor the stored choice.

### DIFF
--- a/src/renderer/src/components/PlayerView.vue
+++ b/src/renderer/src/components/PlayerView.vue
@@ -50,7 +50,6 @@ const anime4kActive = computed(() => webgpuAvailable.value && anime4kPreset.valu
 const showControls = ref(true)
 const showPresetMenu = ref(false)
 const showQualityMenu = ref(false)
-const isStreaming = computed(() => !!activeStreamUrl.value && !activeFilePath.value)
 let controlsTimer: ReturnType<typeof setTimeout> | null = null
 
 // MKV remux state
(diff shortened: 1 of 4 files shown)


src/renderer/src/components/PlayerView.vue
```diff
@@ -657,6 +672,15 @@ function toggleMute(): void {
   video.muted = muted.value
 }
 
+let persistVolumeTimer: ReturnType<typeof setTimeout> | null = null
```
mikkerlo, line 678: Don't forget to clear this timer in `onBeforeUnmount` to prevent memory leaks.

src/renderer/src/components/PlayerView.vue
```diff
@@ -84,6 +83,22 @@ const activeStreamUrl = ref(props.streamUrl)
 const selectedHeight = ref(0)
 const hasQualities = computed(() => props.availableStreams.length > 0)
 
+const isStreaming = computed(() => !!activeStreamUrl.value && !activeFilePath.value)
+const streamingBannerVisible = ref(false)
+let streamingBannerTimer: ReturnType<typeof setTimeout> | null = null
```
mikkerlo, line 88: This timer also needs to be cleared in `onBeforeUnmount`.

src/renderer/src/components/PlayerView.vue
```diff
@@ -657,6 +672,15 @@ function toggleMute(): void {
   video.muted = muted.value
 }
 
+let persistVolumeTimer: ReturnType<typeof setTimeout> | null = null
+watch([volume, muted], ([v, m]) => {
```
mikkerlo, line 680: Consider using `{ immediate: false }` for this watcher or a flag, to avoid writing the volume back to the store immediately after restoring it in `onMounted`.

mikkerlo (commented):
Overall great PR! Volume persistence and auto-translation pick are big UX wins. I have left a few suggestions for cleanup and a minor edge case.

mikkerlo (approved):
LGTM after testing on Windows

PR #31, opened 2026-04-17
feat: offline Shikimori queue + "Working offline" indicator

akvasha:
## Summary
- Intercept transport-level fetch failures in `shikimori:update-rate` (distinguishes `TypeError` from `ShikiApiError`)
- Persist `{ malId, rateId, before, after, queuedAt }` records to `shikimoriUpdateQueue` (electron-store)
- Optimistically update the cached rate so UI reflects the change
- New IPCs: `shikimori:get-offline-queue-length` (invoke) and `shikimori:offline-queue-changed` (broadcast)
- AnimeDetailView shows a compact "Working offline — N changes queued" chip when the queue is non-empty
- Queue is cleared on logout
- Bumps version to 3.8.2

Actual sync-back on reconnect (with conflict resolution against the authoritative Shikimori state) is the next TODO (#3).

## Test plan
- [x] Normal online save: rate updates, no chip appears
- [x] Offline save: save appears successful, chip shows "Working offline — 1 change queued"
- [x] Multiple queued updates increment the counter
- [x] Chip persists across app restart (hydrated via `shikimoriGetOfflineQueueLength`)
- [x] HTTP 4xx error (e.g., invalidated token) propagates to `shikiError`, does NOT queue
- [x] `npm run typecheck` passes

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -40,6 +40,7 @@
 - [x] HEVC (H.265) Support in MSE Streaming Path — `hevcCodecString` produces `hvc1.…` for Main / Main 10 / Main Still Picture; ffmpeg spawn emits `-tag:v hvc1` so Chromium MSE accepts the track; legacy full-remux fallback still fires when the platform has no HEVC decoder
 - [x] HEVC → H.264 transcode fallback for platforms without an HEVC decoder — new `player:remux-mkv-stream-transcode` IPC re-encodes HEVC to H.264 through the existing MSE pipe; `pickH264Encoder` dry-runs `h264_vaapi` / `h264_nvenc` / `h264_qsv` / `libx264` at startup; `hevcTranscodeOnPlay` setting (ask / always / never) with consent modal and `shell:open-external-file` escape hatch
 - [x] Centralized Shikimori Cache & Surgical UI Updates — persist `shikimoriUserRates` in electron-store, cache-first `get-anime-rates`, background API refresh with `shikimori:rates-refreshed` broadcast, surgical `shikimori:rate-updated` on rate changes
+- [x] Offline Shikimori Support: Queuing & Status Indicators — intercept transport-level failures in `shikimori:update-rate`, persist `{before, after}` deltas to `shikimoriUpdateQueue`, optimistically update cache, broadcast `shikimori:offline-queue-changed`, show "Working offline" chip in AnimeDetailView
 
 ---
 
@@ -61,7 +62,7 @@
 - **Schema Design:** Must define a robust schema for `shikimoriUserRates` that includes `updatedAt` to avoid stale cache issues.
 - **IPC Overhead:** Broadcasting full rate objects for large lists (1000+ entries) may cause renderer lag; needs optimized payload or filtered broadcasts.
 
-## 2. Offline Shikimori Support: Queuing & Status Indicators
+## ~~2. Offline Shikimori Support: Queuing & Status Indicators~~ (done)
 
 **Priority:** High | **Effort:** Medium
 
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -84,7 +84,8 @@ const store = new Store({
     anime4kPreset: 'off' as 'off' | 'mode-a' | 'mode-b' | 'mode-c',
     hevcTranscodeOnPlay: 'ask' as 'ask' | 'always' | 'never',
     watchProgress: {} as Record<string, { position: number; duration: number; updatedAt: number; watched?: boolean }>,
-    shikimoriUserRates: [] as unknown[]
+    shikimoriUserRates: [] as unknown[],
+    shikimoriUpdateQueue: [] as unknown[]
   }
 })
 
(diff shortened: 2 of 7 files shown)


src/main/index.ts
```diff
@@ -94,6 +95,21 @@ function broadcastToAll(channel: string, ...args: unknown[]): void {
   }
 }
 
+function isNetworkError(err: unknown): boolean {
+  if (err instanceof shikimori.ShikiApiError) return false
+  if (err instanceof TypeError) return true
```
mikkerlo, line 123: Suggestion: Broaden this check to include common Node.js network error codes like `ENOTFOUND`, `ETIMEDOUT`, `ECONNREFUSED`, and `EAI_AGAIN`. You can check `err.code` for these.

src/main/index.ts
```diff
@@ -1252,35 +1270,77 @@ function registerIpcHandlers(): void {
...
+          }
+          store.set('shikimoriUserRates', cached)
+          broadcastToAll('shikimori:rate-updated', cached[idx])
+        } else {
+          refreshShikimoriRatesInBackground()
+        }
+
+        return updatedRate
+      } catch (err) {
+        if (!isNetworkError(err) || idx === -1) throw err
```
mikkerlo, line 1331: If `idx === -1`, the change is not queued and an error is thrown. This means if a user tries to update an anime that isn't in the currently cached list (e.g. from search results while offline), it will fail entirely. We should ideally support queuing for these cases too, although it would require a placeholder `shikiAnime` object in the cache.

src/main/index.ts
```diff
@@ -1252,35 +1270,77 @@ function registerIpcHandlers(): void {
...
-            target_id: updatedRate.target_id
+            ...cachedEntry.rate,
+            episodes,
+            status,
+            score,
+            updated_at: new Date().toISOString()
           }
         }
         store.set('shikimoriUserRates', cached)
         broadcastToAll('shikimori:rate-updated', cached[idx])
```
mikkerlo, line 1357: The synthetic `ShikiUserRate` returned here is missing the `updated_at` field. While most of our UI uses `shikimori:rate-updated` broadcasts, some components might rely on the return value of the invoke call. It's safer to include `updated_at: new Date().toISOString()` here.

TODO.md
```diff
@@ -61,7 +62,7 @@
 - **Schema Design:** Must define a robust schema for `shikimoriUserRates` that includes `updatedAt` to avoid stale cache issues.
 - **IPC Overhead:** Broadcasting full rate objects for large lists (1000+ entries) may cause renderer lag; needs optimized payload or filtered broadcasts.
```
mikkerlo, line 64: Conflict markers are still present in this file. Please resolve them.

mikkerlo (commented):
Great progress on the offline queuing system! The interception of transport errors and optimistic UI updates are well-thought-out. I have a few suggestions to make the error detection more robust and ensure UI consistency.

LGTM after manual testing.

akvasha:
Addressed the 4 review comments:

1. **`isNetworkError` too narrow** — Fixed. Added `NETWORK_ERROR_CODES` set (`ENOTFOUND`, `ETIMEDOUT`, `ECONNREFUSED`, `ECONNRESET`, `EAI_AGAIN`, `ENETUNREACH`, `EHOSTUNREACH`, `EPIPE`, `UND_ERR_SOCKET`, `UND_ERR_CONNECT_TIMEOUT`) and a recursive `errorCode()` helper that walks `err.cause` chains to catch undici-wrapped Node errors. Existing `TypeError` / `AbortError` / `TimeoutError` branches are preserved.

2. **`idx === -1` skips queueing for anime not in cache** — Keeping as-is for this PR. Rationale: the only callsites are `AnimeDetailView.shikiSave` (requires a resolved MAL ID for anime the user is already viewing) and `PlayerView` auto-tracker (only fires for anime the user has open, which reached us through the rate list or a saved entry). Queuing an update for an anime with no cached `shikiAnime` would require either (a) fetching metadata — which itself needs network and defeats the purpose — or (b) storing a placeholder that the UI would then render as a ghost row. That's a larger scope change best handled alongside TODO #3 (conflict-aware sync) where we can validate the new rate against server state before materializing a cache entry. For now the narrow guard keeps the optimistic-update invariant honest: we only fake success when we have real cached state to roll back to.

3. **Synthetic `ShikiUserRate` missing `updated_at`** — No change needed. The `ShikiUserRate` interface in `src/main/shikimori.ts:24-31` doesn't define an `updated_at` field (that's on `ShikiAnimeRateEntry` — the cache-entry type, not the rate-update return type). The cached entry's `updated_at` **is** updated inside the intercept branch (`cached[idx].rate.updated_at = new Date().toISOString()`), so the broadcast-driven UI path already sees the fresh timestamp. Adding `updated_at` to the synthetic return would make it inconsistent with the online return, which also doesn't include it.

4. **TODO.md conflict markers** — False positive; no markers in the actual file (verified via `grep '<<<\|>>>\|======' TODO.md` → no matches). The diff context shows the before/after of my resolution during the rebase, which github's review tool renders with the tokens still visible.

mikkerlo (approved):
Thank you for checking on the markers! You are correct, they are resolved in this branch. The broader `isNetworkError` check is also much more robust now.

I am approving the PR as the core offline queuing functionality is solid.

### Non-blocking suggestions
1. **Synthetic `updated_at`**: Consider adding `updated_at: new Date().toISOString()` to the return object in the offline branch (line 1341) for complete type/UI consistency.
2. **Uncached Anime**: Queuing updates for anime not currently in the cache (`idx === -1`) could be a nice future improvement.

LGTM after manual testing.

akvasha:
Thanks for the approval! Quick notes on the two non-blocking suggestions so the record is clear:

1. **Synthetic `updated_at`** — keeping as-is. The synthetic return is typed `satisfies shikimori.ShikiUserRate` (`src/main/shikimori.ts:24-31`), and that interface has no `updated_at` field. Adding one would fail the `satisfies` check and would also diverge from the online return path (`createUserRate` / `updateUserRate` cast `response.json()` to `ShikiUserRate` — they don't expose `updated_at` either). I also checked consumers: `shikiRate` in `AnimeDetailView.vue` is only written, never read for `.updated_at` (and PlayerView discards the return). The cached entry's `updated_at` *is* refreshed inside the intercept block (`cached[idx].rate.updated_at = new Date().toISOString()`), which is what the `shikimori:rate-updated` broadcast and the rate list UI actually read.

2. **Queueing for uncached anime** — noting for TODO #3. When we add conflict-aware sync, we'll have a natural point to materialize a placeholder `shikiAnime` from the local catalog (name/image are already in the anime record we navigated from) without a network fetch, then reconcile it against server state on reconnect. Doing it here would either require a network call (defeats the purpose) or a ghost row (rendering edge cases).

PR #32, opened 2026-04-18
feat: conflict-aware automatic sync for offline Shikimori changes

akvasha:
## Summary

Closes TODO #3. Background worker drains `shikimoriUpdateQueue` back to Shikimori when connectivity returns, reconciling each entry against server state so edits made on shikimori.one during the offline window aren't silently clobbered.

- Module-level `syncShikimoriQueue()` with in-memory mutex, 60s timer (runs only while queue non-empty), post-success hooks (after any online `update-rate` or rates refresh), boot kick-off if queue non-empty
- Per-malId consolidation: first `before` + last `after` — one PATCH even if the user bumped episodes multiple times offline
- Conflict rule: if server drifted from `before`, only apply `after` when it's strictly progress (`after.episodes > current.episodes` AND current status not more advanced than `after.status` using `planned < watching/rewatching < on_hold/dropped < completed`). Otherwise reconcile local cache to server and drop the item. Server deletion triggers recreate with `after`
- Outcome-based trigger (same philosophy as TODO #2): no `navigator.onLine` / `net.isOnline`, we test the real fetch result. Items serialized with 250ms spacing to stay under Shikimori's 5 req/s
- 401/403 aborts drain without dropping items (user re-auth needed); other HTTP errors drop the offending item; network errors leave the queue intact for next tick
- New IPCs: `shikimori:get-sync-status`, `shikimori:trigger-sync`, `shikimori:sync-status` broadcast
- AnimeDetailView chip flips amber "Working offline" ↔ blue "Syncing…" with spinner; "Retry now" button while idle; tooltip shows last sync error

Also includes a drive-by fix (separate commit) for the AnimeDetailView initial translation-picker: previously when a user's global type+author default didn't match the downloaded file on disk, `selectedTr` fell through to the streaming option and the Play File button didn't appear. PR 0a346a6 had patched this for next/prev navigation in PlayerView but missed the initial-selection path. Now adds a new Priority 3 between user override and global default: prefer any downloaded translation (same type highest quality first, then any).

## Test plan
- [ ] **Happy path**: offline → bump episodes → online → chip flips Syncing…→ disappears; shikimori.one matches
- [ ] **Non-progress drift**: offline set 5 → set 10 on web → online → drain skips, local reconciles to 10, queue clears
- [ ] **Progress drift**: offline set 15, server has 10 → drain applies 15
- [ ] **Server deletion**: offline edit → delete on web → online → recreate fires
- [ ] **Consolidation**: offline bump 5→6→7 three times → one PATCH only
- [ ] **Retry now** button while offline → momentary Syncing then returns to offline
- [ ] **Boot with queue**: kill offline with queued items, restart online → sync fires
- [ ] **Logout**: timer stops, queue cleared
- [ ] **Downloaded translation fix**: open an anime where your global default differs from the downloaded file's type/author → Play File button visible, downloaded translation pre-selected

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -135,6 +135,225 @@ interface QueuedShikimoriUpdate {
   queuedAt: number
 }
 
+type ShikiSyncState = 'idle' | 'syncing'
+interface ShikiSyncStatus {
+  state: ShikiSyncState
+  queueLength: number
+  lastSyncAt: number
+  lastSyncError: string | null
+}
+
+const SYNC_TIMER_MS = 60_000
+const SYNC_ITEM_DELAY_MS = 250
+
+const STATUS_ORDER: Record<shikimori.ShikiUserRateStatus, number> = {
+  planned: 0,
+  watching: 1,
+  rewatching: 1,
+  on_hold: 2,
+  dropped: 2,
+  completed: 3
+}
+
+let syncInProgress = false
+let syncTimer: NodeJS.Timeout | null = null
+let lastSyncAt = 0
+let lastSyncError: string | null = null
+
+function getQueueLength(): number {
+  return (store.get('shikimoriUpdateQueue') as QueuedShikimoriUpdate[]).length
+}
+
+function getSyncStatus(): ShikiSyncStatus {
+  return {
+    state: syncInProgress ? 'syncing' : 'idle',
+    queueLength: getQueueLength(),
+    lastSyncAt,
+    lastSyncError
+  }
+}
+
+function broadcastSyncStatus(): void {
+  broadcastToAll('shikimori:sync-status', getSyncStatus())
+}
+
+function startSyncTimer(): void {
+  if (syncTimer) return
+  syncTimer = setInterval(() => {
+    void syncShikimoriQueue()
+  }, SYNC_TIMER_MS)
+}
+
+function stopSyncTimer(): void {
+  if (!syncTimer) return
+  clearInterval(syncTimer)
+  syncTimer = null
+}
+
+function adjustSyncTimer(): void {
+  if (getQueueLength() > 0) startSyncTimer()
+  else stopSyncTimer()
+}
+
+function consolidateQueue(queue: QueuedShikimoriUpdate[]): QueuedShikimoriUpdate[] {
+  const map = new Map<number, QueuedShikimoriUpdate>()
+  for (const entry of queue) {
+    const existing = map.get(entry.malId)
+    if (existing) {
+      existing.after = entry.after
+      existing.rateId = entry.rateId ?? existing.rateId
+    } else {
+      map.set(entry.malId, { ...entry })
+    }
+  }
+  return Array.from(map.values())
+}
+
+function sleep(ms: number): Promise<void> {
+  return new Promise((resolve) => setTimeout(resolve, ms))
+}
+
+function reconcileCacheFromRate(malId: number, rate: shikimori.ShikiUserRate): void {
+  const cached = store.get('shikimoriUserRates') as {
+    rate: Record<string, unknown> & { id?: number; target_id: number; episodes: number; status: shikimori.ShikiUserRateStatus; score: number }
+    shikiAnime: unknown
+    smotretAnime: unknown
+  }[]
+  const idx = cached.findIndex((e) => e.rate.target_id === malId)
+  if (idx === -1) return
+  cached[idx] = {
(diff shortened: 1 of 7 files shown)


src/main/index.ts
```diff
@@ -135,6 +135,225 @@ interface QueuedShikimoriUpdate {
...
+  }
+}
+
+function broadcastSyncStatus(): void {
+  broadcastToAll('shikimori:sync-status', getSyncStatus())
+}
+
+function startSyncTimer(): void {
+  if (syncTimer) return
+  syncTimer = setInterval(() => {
```
mikkerlo, line 182: Potential Race Condition: `consolidateQueue` works on a snapshot of the queue, but the queue in `electron-store` might grow while the worker is running. If a new update for `malId` comes in while we are processing its consolidated entry, the subsequent `filter` on line 263 will accidentally drop the new update without it ever being synced. We should probably filter by `queuedAt` or only drop items that were part of the consolidated snapshot.

src/main/index.ts
```diff
@@ -135,6 +135,225 @@ interface QueuedShikimoriUpdate {
...
+  queueLength: number
+  lastSyncAt: number
+  lastSyncError: string | null
+}
+
+const SYNC_TIMER_MS = 60_000
+const SYNC_ITEM_DELAY_MS = 250
+
+const STATUS_ORDER: Record<shikimori.ShikiUserRateStatus, number> = {
+  planned: 0,
```
mikkerlo, line 150: Shikimori treats `watching` and `rewatching` as distinct but equal in 'progress' priority. Your `STATUS_ORDER` correctly maps them to `1`. However, if a user is `rewatching` and an offline update sets them back to `watching`, `shouldApplyOnDrift` might return true if episode counts match. We should decide if `rewatching` -> `watching` is a valid downgrade or just a side-grade.

src/renderer/src/components/AnimeDetailView.vue
```diff
@@ -192,7 +194,17 @@ const episodeRows = computed((): EpisodeRow[] => {
       selectedTr = sorted.find(tr => tr.id === episodeOverrides.value.get(ep.id)) || null
     }
 
-    // Priority 3: Global default (same type + author)
+    // Priority 3: Prefer any downloaded translation on this episode — avoids silent fallback to
+    // streaming when the global default (type+author) doesn't match what's on disk
+    if (!isLocked && !selectedTr && downloadedTrIds.size > 0) {
+      const downloaded = sorted.filter(tr => downloadedTrIds.has(tr.id))
+      const sameType = downloaded
```
mikkerlo: Drive-by fix logic: In the new Priority 3, we prefer same-type downloaded files. If multiple authors/qualities are downloaded for the same type, we pick the highest quality. It might be even better to also check if one of these matches the `selectedAuthor` preference before falling back to just 'highest quality'.

mikkerlo (commented):
Excellent implementation of the conflict-aware sync! The consolidation logic and the progress-based override rule are exactly what we needed to handle server drift safely. The UI feedback in the chip is also a nice touch.

I have a few technical observations regarding the queue consolidation and translation selection logic.

LGTM after manual testing.

akvasha:
Addressed all three review comments in a new commit (not squashed — kept separate per the fix-commit convention):

1. **Consolidation race** — Valid bug. `consolidateQueue` now returns `ConsolidatedWorkItem` objects carrying the exact `consumedQueuedAts` set from the snapshot. The post-processing drop (new helper `dropConsumedEntries`) filters by `(malId, queuedAt)` tuples, so a new offline edit for the same malId that landed while the drain was mid-flight stays in the queue and is picked up on the next tick.

2. **`watching` ↔ `rewatching` side-grade** — Valid edge case. `shouldApplyOnDrift` now rejects same-rank-but-different-value: both are rank `1`, so if current is `rewatching` and after is `watching` (with episode progress), the drift check fires and we reconcile to server rather than downgrading the user's status. Only exact status match or strictly higher rank overrides.

3. **Translation picker author preference** — Fair point. New Priority 3 now prefers `selectedAuthor` within same-type downloaded translations before falling back to highest quality. If the user has multiple authors downloaded for the same type, their stated author preference wins.

Typecheck passes.

mikkerlo (approved):
Excellent work addressing the feedback! The consolidation logic is now much more robust with the timestamp-based filtering, and the refined side-grade rules provide great safety for conflict resolution.

LGTM after manual testing.

PR #33, opened 2026-04-18
feat: gradual background pre-fetching of Shikimori detailed info

akvasha:
## Summary
- Adds a throttled (1 req / 2s) fire-and-forget worker that drains missing/stale `shikimoriAnimeDetails` for the user's Watching + Planned Shikimori list
- Triggers after rate-list refreshes and lazily on `shikimori:get-anime-details` cache misses; aborts on logout; refetches stale entries after 30 days
- Cache is stored separately from rate entries so the heavier details payload doesn't get re-broadcast on every rate update
- AnimeDetailView's Shikimori panel surfaces genres + a collapsible description (bbcode-stripped, with HTML-stripped fallback) when cached, providing usable offline metadata for watchlist entries the user hasn't downloaded
- Implements TODO #4

## Test plan
- [ ] Log in to Shikimori, open an anime in the Watching/Planned list — Shikimori panel shows genres + description within ~2 minutes of first sync
- [ ] Toggle "Show more" / "Show less" on long descriptions
- [ ] Logout clears `shikimoriAnimeDetails` and aborts the worker
- [ ] Anime in Completed status: details are not prefetched; moving to Planned triggers prefetch on next rate refresh
- [ ] Offline: cached anime still renders description; uncached anime renders panel without description (no crash, no spinner)
- [ ] `npm run typecheck` passes


src/main/index.ts
```diff
@@ -85,7 +85,8 @@ const store = new Store({
     hevcTranscodeOnPlay: 'ask' as 'ask' | 'always' | 'never',
     watchProgress: {} as Record<string, { position: number; duration: number; updatedAt: number; watched?: boolean }>,
     shikimoriUserRates: [] as unknown[],
-    shikimoriUpdateQueue: [] as unknown[]
+    shikimoriUpdateQueue: [] as unknown[],
```
mikkerlo, line 88: Suggestion: Add `shikimoriAnimeDetails: {}` to the `Store` defaults to ensure the property is initialized correctly on new installations.

src/main/index.ts
```diff
@@ -160,6 +161,13 @@ let syncTimer: NodeJS.Timeout | null = null
 let lastSyncAt = 0
 let lastSyncError: string | null = null
 
+const PREFETCH_INTER_REQUEST_MS = 2000
```
mikkerlo, line 164: Note: `shikiFetch` already has a 1-second retry for 429 errors. With a 2-second interval here, we are very safe, but remember that this worker's requests consume the same global rate limit as user-initiated actions (like updating rates). If the user is active, this might occasionally trigger the 429 retry in the main thread.

src/renderer/src/components/AnimeDetailView.vue
```diff
@@ -45,6 +45,26 @@ const friendsLoading = ref(false)
...
+  if (src) {
+    return src
+      .replace(/\[\/?[a-zA-Z][^\]]*\]/g, '')
+      .replace(/\s+/g, ' ')
+      .trim()
+  }
+  const html = shikiDetails.value.description_html
+  if (!html) return ''
+  const tmp = document.createElement('div')
+  tmp.innerHTML = html
```
mikkerlo: Performance: Creating a `div` inside a computed property can be slow and technically violates the 'pure function' expectation for computed. Since you already have the BBCode regex, you might want to move the full stripping logic (including HTML) to the main process before caching, or use a more efficient string-only approach here.

mikkerlo (commented):
This is a fantastic addition for offline metadata! The throttled background worker is well-implemented and respects Shikimori's rate limits nicely. I have a few suggestions to improve performance and code safety.

akvasha:
Thanks for the thorough review! Pushed `31499864` addressing each point.

1. **`shikimoriAnimeDetails` in defaults (`src/main/index.ts:88`)** — Already in the defaults block at line 89 (`shikimoriAnimeDetails: {} as Record<string, { details: shikimori.ShikiAnimeDetails; fetchedAt: number }>`). I think the comment was anchored to the line above; no change needed.

2. **Rate-limit budget note (`src/main/index.ts:164`)** — Acknowledged. The 2 s spacing was chosen specifically to leave headroom for concurrent user actions and the offline-sync drain (which runs at 250 ms intervals during its own work). Combined with `shikiFetch`'s built-in 429 retry-after, we should never come close to the 5 req/s ceiling. No code change.

3. **Computed creating a `div` (`AnimeDetailView.vue:64`)** — Fixed. Replaced the `document.createElement('div') / innerHTML` HTML strip with a string-only regex pass (handles `<br>`, generic tags, and the common HTML entities). The computed is now pure and DOM-free.

4–6. **No broadcast on cache write / view stuck on cache miss / lazy fetch returns null** — Fixed together. After a successful `getAnimeDetails`, the worker now broadcasts `shikimori:anime-details-updated` (`{ malId, details }`). `AnimeDetailView` subscribes in `onMounted` and hydrates `shikiDetails` in place when the matching MAL ID arrives, with cleanup in `onUnmounted`. The lazy `shikimori:get-anime-details` cache-miss path stays unchanged (returns `null` immediately, fires the worker), but the broadcast now drives the eventual UI update — no polling, no reopen needed.

7. **DESIGN.md inaccuracy on abort behavior (`DESIGN.md:519`)** — Fixed the doc to match the code: "bails on the next iteration check after the current request and any in-progress 2 s inter-request sleep complete." (Opted for doc accuracy over implementing an abortable sleep, since logout-to-clean-state latency is bounded at ~2 s in the worst case and an abortable sleep adds machinery for negligible UX gain.)

akvasha:
Follow-up: CodeQL flagged two issues on the string-only HTML strip I added in the previous commit. Pushed `97c40229`:

- **Incomplete multi-character sanitization** (line 65) — switched `<[^>]*>` to an iterative strip loop (`do { prev = s; s = s.replace(…, '') } while (s !== prev)`) so nested/malformed patterns like `<scr<script>ipt>` can't slip through.
- **Double escaping / unescaping** (line 67) — consolidated the six sequential entity `.replace()` calls into a single-pass regex with a lookup table (`/&(?:amp|nbsp|lt|gt|quot|#39);/g` + function replacement), so `&amp;lt;` now correctly decodes to the literal text `&lt;` rather than double-unescaping to `<`.

mikkerlo:
I've reviewed your latest updates and responses! You've done an excellent job addressing the technical nuances, especially the CodeQL findings. To ensure my previous comments are perfectly clear, here is a more detailed breakdown of the "why" behind those suggestions:

### 1. Store Defaults (`src/main/index.ts:88`)
- **Explanation**: You correctly identified it was already there at line 89. My concern was about **forward-compatibility**. 
- **The Detail**: When we add new keys to the `Store` defaults, existing users' local `config.json` files won't automatically get the new key unless we explicitly handle a migration. By having it in `defaults`, `electron-store` provides that fallback value automatically for new installs, but for existing users, we need to ensure the code handles an `undefined` return. Your implementation `cache || {}` handles this perfectly.

### 2. Rate-Limit Budget (`src/main/index.ts:164`)
- **Explanation**: Shikimori's API limit is 5 requests per second per IP. 
- **The Risk**: While 2 seconds (0.5 req/s) is very conservative, the **Global** nature of the limit is the key. If the background worker is running AND a user is rapidly clicking through their list or the offline sync worker is draining (which runs at 250ms), we could burst above 5 req/s.
- **The "429" Interaction**: Our `shikiFetch` has a 1s retry. If the background worker triggers a 429, it "stalls" the main thread's fetch loop for 1s. This isn't a crash, but it can make the UI feel slightly less responsive for user-initiated actions. The 2s delay you chose is a great balance.

### 3. DOM Elements in Computed (`AnimeDetailView.vue:64`)
- **Explanation**: Vue's `computed` properties are meant to be "pure" transformations of state. 
- **Performance**: Calling `document.createElement` and setting `innerHTML` triggers the browser's full HTML parser and DOM engine. Doing this inside a computed property—which might re-evaluate frequently during parent re-renders—is significantly more expensive than a regex.
- **Memory**: Every time that computed re-runs, a new detached DOM node is created. While GC eventually cleans them up, your new iterative string-based approach is much lighter and safer for the renderer's main thread.

### 4. Broadcast & UX Improvements (Items 4–7)
- **The "Stuck" UI Problem**: Previously, if a user opened an anime that wasn't in the cache, the IPC would return `null`. The UI would stay empty even after the background worker eventually fetched the data, unless the user closed and reopened the view.
- **Solution**: Your addition of the `shikimori:anime-details-updated` broadcast is the "missing link." It turns a pull-based UI into a reactive one. Now, the view can "hydrate" as soon as the data arrives, which is a much smoother experience.

Your iterative sanitization loop for the HTML strip is a very robust way to handle those nested tag edge-cases. Great work!

LGTM after manual testing.

mikkerlo (approved):

PR #36, opened 2026-04-19
feat: Shikimori series chronology in anime detail view

akvasha:
## Summary

- Add a **Chronology** section to `AnimeDetailView` that surfaces the full franchise graph from Shikimori's `/api/animes/:id/franchise` endpoint.
- Each row shows thumbnail, title, kind badge, year, relation to the current anime (sequel/prequel/…), and — when the user tracks the entry on Shikimori — a watch-status badge (Watched / Watching / Planned / …).
- Clicking a row navigates in place; back button / mouse-back returns via a per-view navigation stack.

## Implementation notes

- **Franchise call is unauthenticated** — Shikimori's `/franchise` endpoint is public, and sending a stale access token makes it respond 400.
- **BFS from `current_id` follows only canonical relations** (sequel, prequel, side_story, parent_story, alternative_version, summary, full_story, spin_off, alternative_setting). Without this, crossover series like *Isekai Quartet* bridge unrelated franchises (Overlord + Re:Zero + Konosuba + Youjo Senki) via `"other"` edges, ballooning the list.
- **Filters out** pv / cm / music nodes (promo content).
- **Kind labels are normalized** from the endpoint's localized Russian kind strings, preferring machine-readable values from the Shikimori details cache + smotret-anime `type` field, falling back to Russian keyword heuristics.
- **Per-view navigation stack** in `App.vue` lets back button pop through a chain of opened anime (A → chronology → B → chronology → C → back → back → back).

## Test plan

- [x] Typecheck passes
- [x] COTE S4 shows S1–S4 chronologically, PV filtered, S1–S3 badged "Watched", S4 "Watching"
- [x] Youjo Senki Movie no longer leaks Overlord / Re:Zero / Konosuba into the list
- [x] Back from a chronology-opened anime returns to the previous anime, not the library
- [x] "Not available" rows render for entries with no smotret-anime match


mikkerlo (dismissed):
### PR Review

**General Feedback**:
Great work! The chronology feature is a fantastic addition. Reusing the `malIdMap` cache to resolve items for deep linking is smart, and the in-place navigation via the per-view history stack in `App.vue` works beautifully.

However, I've noticed a couple of issues that need to be addressed before merging.

**Specific Observations & Suggestions for Improvement**:

**1. Backwards Relation Labels (`src/main/index.ts`)**
Shikimori's franchise graph provides bidirectional links (e.g., `A` is a prequel to `B`, and `B` is a sequel to `A`). In your `relationByNodeId` loop:
```typescript
for (const link of franchise.links) {
  if (link.source_id === franchise.current_id) {
    relationByNodeId.set(link.target_id, link.relation)
  } else if (link.target_id === franchise.current_id) {
    relationByNodeId.set(link.source_id, link.relation)
  }
}
```
The `else if` branch captures the relation from the *target's* perspective relative to the current anime, and since it usually appears later in the `links` array, it overwrites the correct relation. This results in backwards labels in the UI (e.g., a sequel will be labeled as a "parent story" or "prequel"). 
**Action**: You can safely remove the `else if` branch entirely. Shikimori provides all links from the `current_id`'s perspective as `source_id`, so `if (link.source_id === franchise.current_id)` is sufficient.

**2. Chronology is accidentally hidden for unauthenticated users (`src/renderer/src/components/AnimeDetailView.vue`)**
As you mentioned in the PR notes, the `/franchise` call is unauthenticated. However, the Chronology is currently completely hidden for users who aren't logged into Shikimori.
- In `loadShikimoriData()`, there is an early return (`if (!shikiUser.value || !anime.value?.myAnimeListId) return`) right before your new code. So `shikimoriGetRelated` is never called for unauthenticated users.
- The template uses `v-if="... && (shikiUser || !shikiUserChecked) ..."` which actively hides the panel when logged out.
**Action**: Move the `relatedPromise` fetch block above the early return in `loadShikimoriData()` and remove the `(shikiUser || !shikiUserChecked)` condition from the `related-panel`'s `v-if` so everyone can see the chronology!

Please make these adjustments, and this will be good to go!

mikkerlo:
Also let's make chronology hidden by default

akvasha:
Thanks for the review! Pushed a follow-up commit with fixes for everything raised:

**mikkerlo**
1. **Backwards relation labels** — Dropped the `else if` branch in `relationByNodeId`. You were right: Shikimori emits every current-node edge as `source_id === current_id`, so the reverse branch only overwrote correct labels with their inverse.
2. **Chronology hidden when logged out** — Moved `relatedPromise` above the `!shikiUser` early return in `loadShikimoriData()` and removed the `(shikiUser || !shikiUserChecked)` gate from the `related-panel` `v-if`. Watch-status badges stay conditional on `watchStatus`, so they naturally disappear for logged-out users.

**Copilot**
1. **BFS `queue.shift()` O(n)** — Switched to an index cursor (`for (let i = 0; i < queue.length; i++)`) to keep traversal linear.
2. **DESIGN.md IPC table** — Fixed endpoint name `/api/animes/:id/related` → `/api/animes/:id/franchise`.
3. **DESIGN.md Series Chronology paragraph** — Rewrote to match the implementation: BFS canonical-relation filter (with the Isekai-Quartet example), per-view navigation stack in `App.vue`, chevron-collapse (not a "Show more" toggle), and that the panel renders for logged-out users since `/franchise` is public.
4, 5. Same as mikkerlo #2 — covered above.
6. **Accessibility** — Added `role="button"`, `tabindex`, and `@keydown.enter`/`@keydown.space` handlers to clickable chronology rows; added `alt` on the thumbnail; added `:focus-visible` styling so keyboard focus is visible.

akvasha:
Round 2: rebased on main (resolved a `package.json` version conflict — main is on 3.11.4 now, so bumped this PR to 3.12.0 for the new feature) and added a follow-up commit that flips `relatedCollapsed` to start `true` on every mount, so the Chronology panel now opens collapsed regardless of entry count. Dropped the post-fetch `> 8` re-derivation and updated the DESIGN.md wording accordingly.

mikkerlo (approved):
### PR Review

**General Feedback**:
This is a fantastic update. Thanks for addressing both the reverse relation bug and the unauthenticated user visibility issue so quickly. The new keyboard accessibility attributes (`role="button"`, `tabindex="0"`) and focus outlines are excellent touches that significantly improve the UX. Collapsing the chronology panel by default is also a great call to keep the initial page load clean.

**Specific Observations**:
- Removing the reverse link edge mapping successfully fixes the relation bug.
- Re-ordering the early returns and removing the `v-if` authentication check properly exposes the feature to all users.
- Nice addition of the A11y roles and handlers.

Awesome work!

LGTM after manual testing

PR #37, opened 2026-04-22
Watch Together (Syncplay)

mikkerlo:
## Summary

Implements Watch Together via the Syncplay protocol. Compatible with the public `syncplay.pl` servers and the reference mpv/VLC Syncplay client, so a friend on any Syncplay-supported player can watch along.

> [!WARNING]
> **Feature in development.** Sync, attribution, and reconnect behavior may misfire — especially on rapid arrow-key seeks or shaky connections. The Settings tab is marked accordingly.

- **Main protocol client** (`src/main/syncplay.ts`): TCP/TLS with Hello handshake, STARTTLS upgrade, line-delimited JSON parser, `ignoringOnTheFly` echo-suppression counters, 1 s heartbeat, RTT compensation, exponential-backoff auto-reconnect.
- **File identity**: canonical `"{animeName} - {episodeInt}"` + duration so instances on different translations/qualities/local-vs-CDN still match; `features.animeDlAppMeta = {animeId, malId, episodeInt, translationId}` rides along so two instances of this app auto-navigate when the other advances episodes.
- **Readiness gate**: when either side's MSE buffer runs out (`waiting` event, debounced 600 ms; or the custom `waitForBufferAhead` path), `Set.ready(false)` keeps both clients paused until everyone's caught up. Member list shows a green/amber dot per user.
- **"Paused by X" badge** in the player while paused; only updates on actual play↔pause transitions to avoid mis-attribution from heartbeats.
- **Settings** → new Watch Together tab (host, port, room, username, password, preferTls, autoReconnect, Test connection). Tab carries an "in development" badge. Persisted under new `syncplay` electron-store key; session state not persisted across restarts.
- **PlayerView** → Syncplay popover button in controls bar, local video events → remote state, remote state applied with 3 s drift tolerance (or `doSeek`), echo suppression via main-side counter + 1.5 s renderer flag, 1 s renderer-pushed position snapshot.
- **Remote episode auto-nav** when `animeId` matches current anime; toast-only for cross-anime or missing episodes.
- `DESIGN.md` updated (Watch Together section, IPC rows, `syncplay:set-ready` row, readiness gate description).
- Rebased onto current `main` (v3.17.0); version bumped to **3.18.0**.

Debug tracing behind `SYNCPLAY_DEBUG=1` env var, plus `syncplay:trace` IPC for renderer-side console logging of every inbound/outbound message.

## Test plan

- [ ] Two instances of this app, same room, same episode — play/pause/seek sync within ~500 ms
- [ ] One instance on CDN stream, other on local MKV of same episode — still syncs on position
- [ ] Seek into unbuffered region on one side — both stay paused via readiness gate, resume together
- [ ] Pause on A — B sees "Paused by A", A sees "Paused by you"; attribution survives 1 s heartbeat
- [ ] Rapid arrow-key seeks — no pause/play ping-pong (still flaky, see warning above)
- [ ] User A advances to next episode — User B auto-navigates
- [ ] User A switches translation mid-episode — position sync continues
- [ ] Interop with mpv + Syncplay reference client in the same room — bidirectional
- [ ] Network drop — toast, auto-reconnect, resume
- [ ] Wrong password — error toast, no retry storm
- [ ] Settings > Watch Together > Test connection — reports OK or the server error

### DIFF
--- a/src/main/syncplay.ts
+++ b/src/main/syncplay.ts
@@ -1,0 +1,724 @@
+import * as net from 'net'
+import * as tls from 'tls'
+import { EventEmitter } from 'events'
+
+const CLIENT_VERSION = '1.6.9'
+const HEARTBEAT_MS = 1000
+const MAX_RECONNECT_ATTEMPTS = 5
+const RECONNECT_BASE_MS = 1000
+const DEBUG = process.env.SYNCPLAY_DEBUG === '1' || process.env.SYNCPLAY_DEBUG === 'true'
+
+function log(...args: unknown[]): void {
+  if (DEBUG) console.log('[syncplay]', ...args)
+}
+
+export interface SyncplayConfig {
+  host: string
+  port: number
+  room: string
+  username: string
+  password?: string
+  autoReconnect: boolean
+}
+
+export type SyncplayState =
+  | 'idle'
+  | 'connecting'
+  | 'tls-probing'
+  | 'tls-handshake'
+  | 'hello-sent'
+  | 'ready'
+  | 'reconnecting'
+  | 'disconnected'
+
+export interface SyncplayStatus {
+  state: SyncplayState
+  host?: string
+  port?: number
+  room?: string
+  username?: string
+  tls?: boolean
+  error?: string
+}
+
+export interface SyncplaySnapshot {
+  position: number
+  paused: boolean
+}
+
+export interface SyncplayFileInfo {
+  animeId: number
+  malId: number | null
+  episodeInt: string
+  translationId: number | null
+  canonicalName: string
+  duration: number
+}
+
+export interface SyncplayRemoteState {
+  paused: boolean
+  position: number
+  setBy: string | null
+  doSeek: boolean
+}
+
+export interface SyncplayRoomUser {
+  username: string
+  file: { name: string; duration: number; size?: number } | null
+  isReady?: boolean
+  animeDlAppMeta?: {
+    animeId: number
+    malId: number | null
+    episodeInt: string
+    translationId: number | null
+  }
+}
+
+export interface SyncplayRemoteEpisode {
+  animeId: number
+  malId: number | null
+  episodeInt: string
+  translationId: number | null
+  canonicalName: string
+  fromUser: string
+}
+
+export interface SyncplayRoomEvent {
+  level: 'info' | 'warn' | 'error' | 'chat'
+  text: string
+}
+
+type JsonObject = Record<string, unknown>
+
+function isObject(v: unknown): v is JsonObject {
+  return typeof v === 'object' && v !== null && !Array.isArray(v)
+}
+
+export class SyncplayClient extends EventEmitter {
+  private config: SyncplayConfig | null = null
+  private socket: net.Socket | tls.TLSSocket | null = null
+  private rxBuffer = ''
+  private status: SyncplayStatus = { state: 'idle' }
+
+  private heartbeatTimer: NodeJS.Timeout | null = null
+  private reconnectTimer: NodeJS.Timeout | null = null
+  private reconnectAttempts = 0
+
+  private snapshot: SyncplaySnapshot = { position: 0, paused: true }
+  private currentFile: SyncplayFileInfo | null = null
+  private serverMotd = ''
+  private tlsUpgraded = false
+  private roomUsers: SyncplayRoomUser[] = []
+  private ownIsReady = true
+
+  private clientIgnoreCounter = 0
+  private pendingClientAck = 0
+  private pendingServerAck = 0
+  private lastAppliedRoomEpisode: string | null = null
+
+  private serverRtt = 0
+
+  connect(config: SyncplayConfig): void {
+    this.disconnectInternal(false)
+    this.config = config
+    this.reconnectAttempts = 0
+    this.lastAppliedRoomEpisode = null
+    this.openSocket()
+  }
+
+  disconnect(): void {
+    this.disconnectInternal(true)
+  }
+
+  getStatus(): SyncplayStatus {
+    return { ...this.status }
+  }
+
+  setFile(file: SyncplayFileInfo): void {
+    this.currentFile = file
+    if (this.status.state === 'ready') this.sendSetFile(file)
+  }
+
+  setReady(isReady: boolean): void {
+    this.ownIsReady = isReady
+    if (this.status.state === 'ready') this.sendSetReady(isReady)
+    this.updateOwnReadinessInRoom()
+  }
+
+  private updateOwnReadinessInRoom(): void {
+    if (!this.config) return
+    const me = this.roomUsers.find((u) => u.username === this.config!.username)
+    if (me) {
+      if (me.isReady === this.ownIsReady) return
+      me.isReady = this.ownIsReady
+    } else {
+      this.roomUsers.push({ username: this.config.username, file: null, isReady: this.ownIsReady })
+    }
+    this.emit('room-users', this.roomUsers.slice())
+  }
+
+  sendLocalState(payload: {
+    paused: boolean
+    position: number
+    cause: 'play' | 'pause' | 'seek'
+  }): void {
+    this.snapshot = { position: payload.position, paused: payload.paused }
+    this.clientIgnoreCounter += 1
+    this.pendingClientAck = this.clientIgnoreCounter
+    log('local-state', payload.cause, 'counter=', this.clientIgnoreCounter, 'pos=', payload.position)
+    this.sendStateMessage({ doSeek: payload.cause === 'seek' })
+  }
+
+  updateSnapshot(snap: SyncplaySnapshot): void {
+    this.snapshot = snap
+  }
+
+  private tearDown(): void {
+    if (this.reconnectTimer) {
+      clearTimeout(this.reconnectTimer)
+      this.reconnectTimer = null
+    }
+    this.stopHeartbeat()
+    if (this.socket) {
+      try {
+        this.socket.removeAllListeners()
+        this.socket.destroy()
+      } catch {
+        // ignore
+      }
+      this.socket = null
+    }
+    this.rxBuffer = ''
+    this.tlsUpgraded = false
+    this.clientIgnoreCounter = 0
+    this.pendingClientAck = 0
+    this.pendingServerAck = 0
+    this.serverRtt = 0
+    this.roomUsers = []
+    this.ownIsReady = true
+  }
+
+  private disconnectInternal(userInitiated: boolean): void {
+    this.tearDown()
+    if (userInitiated) {
+      this.config = null
+      this.setStatus({ state: 'idle' })
+    }
+  }
+
+  private openSocket(): void {
+    if (!this.config) return
+    const { host, port } = this.config
+    this.setStatus({
+      state: this.reconnectAttempts > 0 ? 'reconnecting' : 'connecting',
+      host,
+      port,
+      room: this.config.room,
+      username: this.config.username,
+      tls: false
+    })
+    log('connecting', host, port)
+
+    const sock = net.createConnection({ host, port })
+    sock.setKeepAlive(true, 30_000)
+    this.socket = sock
+    this.rxBuffer = ''
+
+    sock.on('connect', () => {
+      log('tcp connected — probing TLS')
+      this.setStatus({ state: 'tls-probing' })
+      // TLS-only: probe before sending Hello so credentials are never on the
+      // wire in plaintext. The server replies with {TLS:{startTLS:'true'|'false'}};
+      // anything other than 'true' aborts the connection.
+      this.sendJson({ TLS: { option: 'send' } })
+    })
+    sock.on('data', (chunk) => this.onData(chunk))
+    sock.on('error', (err) => this.onSocketError(err))
+    sock.on('close', () => this.onSocketClose())
+  }
+
+  private onData(chunk: Buffer | string): void {
+    this.rxBuffer += typeof chunk === 'string' ? chunk : chunk.toString('utf8')
+    let idx: number
+    while ((idx = this.rxBuffer.indexOf('\n')) >= 0) {
+      const line = this.rxBuffer.slice(0, idx).trim()
+      this.rxBuffer = this.rxBuffer.slice(idx + 1)
+      if (!line) continue
+      let msg: unknown
+      try {
+        msg = JSON.parse(line)
+      } catch (err) {
+        log('json parse error', err, line.slice(0, 200))
+        continue
+      }
+      if (!isObject(msg)) continue
+      this.dispatch(msg)
+    }
+  }
+
+  private dispatch(msg: JsonObject): void {
+    const keys = Object.keys(msg).join(',')
+    log('<<', keys)
+    if (DEBUG) this.emit('trace', { dir: 'in', keys, msg })
+    if ('Hello' in msg) this.handleHello(msg.Hello)
+    if ('Set' in msg) this.handleSet(msg.Set)
+    if ('List' in msg) this.handleList(msg.List)
+    if ('State' in msg) this.handleState(msg.State)
+    if ('Chat' in msg) this.handleChat(msg.Chat)
+    if ('Error' in msg) this.handleError(msg.Error)
+    if ('TLS' in msg) this.handleTls(msg.TLS)
+  }
+
+  private handleHello(payload: unknown): void {
+    if (!isObject(payload) || !this.config) return
+    const motd = typeof payload.motd === 'string' ? payload.motd : ''
+    this.serverMotd = motd
+
+    const advertisedUsername = typeof payload.username === 'string' ? payload.username : this.config.username
+    if (advertisedUsername !== this.config.username) {
+      this.emit('room-event', {
+        level: 'info',
+        text: `Joined as "${advertisedUsername}" (server rewrote username)`
+      })
+      this.config = { ...this.config, username: advertisedUsername }
+    }
+
+    if (motd) this.emit('room-event', { level: 'info', text: motd })
+
+    this.finishHandshake()
+  }
+
+  private finishHandshake(): void {
+    if (!this.config) return
+    this.setStatus({
+      state: 'ready',
+      host: this.config.host,
+      port: this.config.port,
+      room: this.config.room,
+      username: this.config.username,
+      tls: this.tlsUpgraded
+    })
+    this.reconnectAttempts = 0
+    if (this.currentFile) this.sendSetFile(this.currentFile)
+    this.sendSetReady(this.ownIsReady)
+    this.updateOwnReadinessInRoom()
+    this.startHeartbeat()
+  }
+
+  private handleTls(payload: unknown): void {
+    if (!isObject(payload)) return
+    if (this.tlsUpgraded) {
+      // Spurious post-upgrade TLS message — ignore.
+      return
+    }
+    if (payload.startTLS === 'true' || payload.startTLS === true) {
+      this.upgradeToTls()
+    } else {
+      this.failHandshake('Server does not support TLS — TLS-only mode requires a TLS-capable Syncplay server (1.6.3+).')
+    }
+  }
+
+  private upgradeToTls(): void {
+    if (!this.socket || !this.config) return
+    log('upgrading to TLS')
+    this.setStatus({ state: 'tls-handshake' })
+    const raw = this.socket
+    raw.removeAllListeners('data')
+    raw.removeAllListeners('error')
+    raw.removeAllListeners('close')
+    const tlsSock = tls.connect({
+      socket: raw,
+      servername: this.config.host
+    })
+    this.socket = tlsSock
+    this.rxBuffer = ''
+    tlsSock.on('secureConnect', () => {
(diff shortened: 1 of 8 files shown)


mikkerlo:
Let's build syncplay's certificates in the app. And use TLS by default

mikkerlo (commented):
Awesome work on implementing the Watch Together feature!

### General Feedback
This is a massive addition to the app that successfully integrates a fairly complex external protocol (Syncplay) without muddying the existing UI. The architecture respects the project's established IPC patterns, keeping protocol logic cleanly separated in `src/main/syncplay.ts`.

### Specific Observations
- **`ignoringOnTheFly` Echo Suppression:** The combination of the main-side counter round-trip checking and the renderer-side 250ms suppression flag is very solid. This avoids the dreaded play/pause ping-pong.
- **Resource Cleanup:** Great job ensuring `syncplaySnapshotTimer` and `syncplayToastTimer` are cleared in `PlayerView.vue`'s `onBeforeUnmount`, preventing memory leaks and zombie intervals.
- **File Identity & Auto-Nav:** Piggybacking `animeDlAppMeta` in the Syncplay `features` dictionary is a brilliant way to allow instances to stay synced and auto-navigate without requiring a custom Syncplay server.
- **UI:** The settings tab is straightforward, and the in-player controls match the existing player overlay style nicely.

- In `SettingsView.vue`, the `syncplaySaveTimer` is not cleared when the component unmounts. While not a huge deal since the settings view stays mounted or is rarely rapidly toggled, clearing it in `onUnmount` ensures no pending saves trigger on a destroyed component.
- `window.api.offSyncplayConnectionStatus()` in `SettingsView.vue` removes all listeners globally (due to `ipcRenderer.removeAllListeners`). Given the app's current view lifecycle, this is safe, but passing specific listener functions to `ipcRenderer.removeListener` in the future could be more resilient.

LGTM after manual testing!

src/main/syncplay.ts
```diff
@@ -0,0 +1,713 @@
...
+      }
+      if (!isObject(msg)) continue
+      this.dispatch(msg)
+    }
+  }
+
+  private dispatch(msg: JsonObject): void {
+    const keys = Object.keys(msg).join(',')
+    log('<<', keys)
+    this.emit('trace', { dir: 'in', keys, msg })
```
akvasha: PR description says debug tracing is gated behind `SYNCPLAY_DEBUG=1`, but `emit('trace', …)` fires unconditionally on every inbound message. Combined with `PlayerView.vue:2016` (`onSyncplayTrace` → `console.log` every entry) and `index.ts:4182` (`broadcastToAll('syncplay:trace', …)`), the renderer console gets ~2 IPC events/sec even at idle — outbound 1 Hz heartbeat + inbound `State` echo. Same fix needed at line 616 (outbound).

```suggestion
    if (DEBUG) this.emit('trace', { dir: 'in', keys, msg })
```

src/main/syncplay.ts
```diff
@@ -0,0 +1,713 @@
...
+    this.sendJson({ State: msg })
+  }
+
+  private sendJson(obj: unknown): void {
+    if (!this.socket) return
+    const keys = Object.keys(obj as JsonObject).join(',')
+    try {
+      this.socket.write(JSON.stringify(obj) + '\r\n')
+      log('>>', keys)
+      this.emit('trace', { dir: 'out', keys, msg: obj })
```
akvasha: Outbound trace also needs the `DEBUG` gate (see comment on line 258).

```suggestion
      if (DEBUG) this.emit('trace', { dir: 'out', keys, msg: obj })
```

src/main/syncplay.ts
```diff
@@ -0,0 +1,713 @@
...
+      this.roomUsers.push(user)
+    }
+    user.file = { name, duration, size }
+    user.animeDlAppMeta = meta
+
+    this.emit('room-users', this.roomUsers.slice())
+    this.emit('room-event', {
+      level: 'info',
+      text: `${username} switched to "${name}"`
+    })
```
akvasha: Syncplay echoes `Set.file` back to the sender, so this `room-event` fires for our own user too — every `pushSyncplayFile()` will surface a `"<your-username> switched to ..."` toast in the player. The `remote-episode-change` emission immediately below already guards against self via `username !== this.config?.username`; the toast wants the same guard.

```suggestion
    this.emit('room-users', this.roomUsers.slice())
    if (username !== this.config?.username) {
      this.emit('room-event', {
        level: 'info',
        text: `${username} switched to "${name}"`
      })
    }
```

src/renderer/src/components/PlayerView.vue
```diff
@@ -1899,6 +2190,20 @@ onBeforeUnmount(() => {
...
+  window.api.offSyncplayRemoteEpisodeChange()
+  window.api.offSyncplayTrace()
+  if (syncplaySnapshotTimer) {
+    clearInterval(syncplaySnapshotTimer)
+    syncplaySnapshotTimer = null
+  }
+  if (syncplayToastTimer) {
+    clearTimeout(syncplayToastTimer)
+    syncplayToastTimer = null
+  }
```
akvasha, line 2246: `syncplayWaitingTimer` is set in `onVideoWaiting` (debounced 600 ms) but never cleared on unmount. If the user navigates away mid-debounce, the timer fires post-teardown and calls `setSyncplayLocalReady(false)` on a torn-down component — matches the timer-cleanup pitfall called out in the project's review checklist.

```suggestion
  if (syncplayToastTimer) {
    clearTimeout(syncplayToastTimer)
    syncplayToastTimer = null
  }
  if (syncplayWaitingTimer) {
    clearTimeout(syncplayWaitingTimer)
    syncplayWaitingTimer = null
  }
```

src/renderer/src/components/SettingsView.vue
```diff
@@ -790,6 +819,78 @@ watch(playerMode, (val) => { if (loaded.value) autoSave('playerMode', val) })
...
+      window.api.syncplayDisconnect()
+    } else if (status.state === 'disconnected') {
+      if (syncplayTestStatus.value === 'testing') {
+        syncplayTestStatus.value = 'failed'
+        syncplayTestError.value = status.error || 'Connection closed'
+      }
+      window.api.offSyncplayConnectionStatus()
+    }
+  }
+  window.api.offSyncplayConnectionStatus()
```
akvasha: `offSyncplayConnectionStatus` in preload is `ipcRenderer.removeAllListeners('syncplay:connection-status')` — clicking "Test connection" while `PlayerView.vue` is mounted (e.g. user has the player open in another tab) will rip out PlayerView's listener too, so PlayerView stops receiving connection-status updates after the test. Same risk on lines 858, 865, 884, 891.

Fix shape: change preload's `on*`/`off*` pair to keep a wrapper-callback map and expose `offSyncplayConnectionStatus(handler)` that does `ipcRenderer.removeListener('syncplay:connection-status', wrappedFor[handler])`. Then this test handler can be torn down without touching PlayerView's. Same pattern applies to all the other Syncplay `off*` helpers.

src/main/syncplay.ts
```diff
@@ -0,0 +1,713 @@
...
+    tlsSock.on('data', (chunk) => this.onData(chunk))
+    tlsSock.on('error', (err) => this.onSocketError(err))
+    tlsSock.on('close', () => this.onSocketClose())
+  }
+
+  private failHandshake(reason: string): void {
+    log('handshake failed:', reason)
+    this.emit('room-event', { level: 'error', text: reason })
+    this.setStatus({ state: 'disconnected', error: reason })
+    this.disconnectInternal(true)
```
akvasha: nit: `failHandshake` emits `setStatus({state:'disconnected', error: reason})` immediately followed by `disconnectInternal(true)` which emits `setStatus({state:'idle'})`. Net effect on listeners: two `connection-status` broadcasts in quick succession, and the `error` field gets clobbered before consumers can read it (Settings → Test connection only branches on `'ready'` / `'disconnected'`, so it currently misses the disconnected event entirely if `disconnectInternal` runs first... wait, here it does run last — but a defensive consumer expecting an idle state will see `error` even though config was reset).

src/renderer/src/components/PlayerView.vue
```diff
@@ -1765,6 +1967,95 @@ onMounted(async () => {
...
+      syncplayLastRemotePlaying = false
+      syncplayLastAppliedPaused = null
+      syncplayPausedBy.value = null
+      if (syncplayWaitingTimer) {
+        clearTimeout(syncplayWaitingTimer)
+        syncplayWaitingTimer = null
+      }
+    }
+    if (status.state === 'reconnecting') {
+      showSyncplayToast('Reconnecting to Syncplay server…', 8000)
```
akvasha, line 2010: nit: `syncplayLastRemotePlaying` survives an episode switch (only `pushSyncplayFile()` runs in the watcher). After a manual prev/next while a remote peer was playing, the new episode auto-plays as soon as `applySyncplayReadyGate` is called — could be intentional (binge-through) but worth a comment so someone tweaking this later doesn't accidentally invert the behavior.

akvasha (commented):
Solid first cut for an in-development feature — TLS-only handshake, RTT-compensated apply rule, `ignoringOnTheFly` bookkeeping, and the readiness gate are all wired. A few fixable issues called out inline before this leaves "in development": (1) trace IPC fires unconditionally despite `SYNCPLAY_DEBUG=1` gating; (2) self-echoed `Set.file` produces a self-toast; (3) `syncplayWaitingTimer` missing from `onBeforeUnmount`; (4) Settings → Test connection's `offSyncplayConnectionStatus()` removes PlayerView's listener too because preload uses `removeAllListeners`.

src/main/syncplay.ts
mikkerlo: Done in 753e1f25 — `emit('trace', …)` is now gated by the `DEBUG` flag at both the inbound dispatch (line 258) and outbound `sendJson` (line 616). At idle, the renderer console no longer sees the 1 Hz heartbeat noise unless `SYNCPLAY_DEBUG=1` is set.

src/main/syncplay.ts
mikkerlo: Done in 753e1f25 — outbound trace at line 616 is now also gated by `DEBUG`.

src/main/syncplay.ts
mikkerlo: Fixed in 753e1f25 — `absorbRemoteFile` now wraps the "X switched to ..." toast in the same `username !== this.config?.username` guard that `remote-episode-change` already used, so the sender doesn't see a self-toast on the server's echo of `Set.file`.

src/renderer/src/components/PlayerView.vue
mikkerlo: Fixed in 753e1f25 — `syncplayWaitingTimer` is now cleared in PlayerView's `onBeforeUnmount` (alongside the existing `syncplaySnapshotTimer` / `syncplayToastTimer` cleanups).

src/renderer/src/components/SettingsView.vue
mikkerlo: Fixed in 753e1f25 — preload now keeps a per-channel `Map<callback, wrapped>` and exposes `offSyncplay*(handler)` that calls `ipcRenderer.removeListener(channel, wrapped)` for the specific handler. SettingsView's "Test connection" passes its local `handler` to `offSyncplayConnectionStatus(handler)` so PlayerView's listener is left in place. PlayerView captures its callbacks at register time (`syncplay*Handler` module-scope vars) and tears them down individually in `onBeforeUnmount`.

src/main/syncplay.ts
mikkerlo: Fixed in 753e1f25 — extracted a private `tearDown()` helper that owns socket / timer cleanup. `failHandshake` now calls `tearDown()` and emits exactly one `setStatus({ state: 'disconnected', error: reason })`, so listeners see a single broadcast carrying the error instead of disconnected-with-error followed by idle clobbering it. `disconnectInternal` keeps the same external behavior by wrapping `tearDown()` + the `idle` broadcast.

src/renderer/src/components/PlayerView.vue
mikkerlo: Fixed in 753e1f25 — added a comment on the `watch([activeEpisodeIndex, activeTranslationId], () => pushSyncplayFile())` block explicitly stating that `syncplayLastRemotePlaying` is intentionally preserved across episode/translation switches so the gate auto-resumes after buffer fills (binge-through). Hopefully that warns the next person off accidentally inverting it.

mikkerlo:
@mikkerlo @akvasha thanks for the reviews. Inline replies posted on each thread. Quick summary of what changed in 753e1f25:

- **trace IPC gating** — `SYNCPLAY_DEBUG=1` now actually controls whether `emit('trace', …)` fires (both inbound & outbound). Renderer console is silent by default again.
- **Self-toast on `Set.file`** — server-echoed file change for our own user is now suppressed.
- **`syncplayWaitingTimer` & `syncplaySaveTimer`** — cleared in PlayerView's `onBeforeUnmount` and SettingsView's `onUnmounted` respectively (matches the existing `syncplaySnapshotTimer`/`syncplayToastTimer` cleanup pattern).
- **Per-listener `off*` helpers** — preload keeps a `Map<callback, wrapped>` per channel; `offSyncplay*(handler)` calls `ipcRenderer.removeListener` for that specific wrapped fn instead of `removeAllListeners`. PlayerView captures its handlers at register-time and tears them down individually; SettingsView's Test connection passes its local handler. So the test connection no longer rips out PlayerView's listeners if the player is open.
- **`failHandshake` double broadcast** — extracted a private `tearDown()`; the failure path now emits exactly one `disconnected` event with the error (no trailing `idle` clobber).
- **`syncplayLastRemotePlaying` preserved across episode switch** — kept as-is (binge-through is intentional) but added a comment so future-me doesn't break it.

CodeQL alert is also resolved via the previous commit (`c3c720c`): TLS handshake is now mandatory and cert validation uses Node's default CA bundle.

akvasha (approved):
All seven previous-review comments addressed in `753e1f25` — verified by reading the file at the head SHA. CI is green across Linux/Windows/macOS + CodeQL. LGTM after manual testing per the PR's test plan (two-instance sync, mpv interop, network-drop reconnect).

akvasha (commented):
Second-pass review — five more findings on top of the first round, plus one parallel bug to the `failHandshake` fix that landed last time but wasn't applied to its sibling site. The Settings → Test connection / live-PlayerView interaction (#1 below) is the one that's most user-visible.

src/renderer/src/components/SettingsView.vue
```diff
@@ -790,6 +823,77 @@ watch(playerMode, (val) => { if (loaded.value) autoSave('playerMode', val) })
...
+        syncplayTestStatus.value = 'failed'
+        syncplayTestError.value = status.error || 'Connection closed'
+      }
+      window.api.offSyncplayConnectionStatus(handler)
+    }
+  }
+  window.api.onSyncplayConnectionStatus(handler)
+
+  try {
+    await window.api.syncplayConnect({
```
akvasha, line 875: Clicking "Test connection" while PlayerView is already connected silently kills the active session. `syncplay.connect()` (in `syncplay.ts:121`) starts with `disconnectInternal(false)` — it tears down whatever connection currently exists before opening the test socket. After the test, this handler calls `syncplayDisconnect()`. Net effect: any user testing settings while watching together drops their friend mid-episode without warning.

Fix shapes: either (a) refuse to test if `syncplayGetStatus()` returns a state other than `idle` / `disconnected`, with a hint to disconnect first; or (b) restore the previous session in `finally` (re-call `syncplayConnect` with the prior config). (a) is simpler.

src/main/syncplay.ts
```diff
@@ -0,0 +1,724 @@
...
+        this.setStatus({
+          state: 'disconnected',
+          host: prevConfig.host,
+          port: prevConfig.port,
+          room: prevConfig.room,
+          username: prevConfig.username,
+          error: text
+        })
+      }
+    }
```
akvasha, line 552: Same double-broadcast pattern that was fixed in `failHandshake` last round. `disconnectInternal(true)` emits `setStatus({state:'idle'})`, then this handler immediately overrides with `disconnected`. Order of broadcasts: `idle` → `disconnected`, and any consumer that synchronously reads status after the `idle` event (e.g. the SettingsView test handler, line 864 `if (status.state === 'disconnected')`) misses the disconnected event entirely or sees the wrong state.

Match the `failHandshake` shape: `tearDown(); this.config = null; setStatus({state:'disconnected', host, port, room, username, error: text})`.

src/renderer/src/components/PlayerView.vue
```diff
@@ -1765,6 +1982,102 @@ onMounted(async () => {
...
+    if (idx === activeEpisodeIndex.value) return
+    showSyncplayToast(`${ep.fromUser} moved to episode ${ep.episodeInt}`)
+    const dir = idx > activeEpisodeIndex.value ? 'next' : 'prev'
+    // goToEpisode moves one step; step toward target in a loop.
+    const stepTowards = async (): Promise<void> => {
+      while (activeEpisodeIndex.value !== idx && !navigating.value) {
+        await goToEpisode(dir)
+      }
+    }
+    stepTowards()
```
akvasha, line 2060: `stepTowards()` is a fire-and-forget async function with three failure modes:

1. No `.catch()` — if `goToEpisode` rejects (e.g. translation fetch fails), it surfaces as an unhandled rejection.
2. No exit guard if `goToEpisode` doesn't advance `activeEpisodeIndex` (e.g. boundary of episode list, or the next episode has no playable translation). The loop runs forever, hammering `goToEpisode` until either `navigating.value` flips or the component unmounts.
3. No cancellation on `onBeforeUnmount`, so user closing the player mid-step leaves the loop running and mutating state on a torn-down component.

Minimum fix: wrap in `try/catch`, add `let safety = props.allEpisodes.length` and decrement each iteration, and capture a local `cancelled` flag that `onBeforeUnmount` sets to true. Plus a module-level mutex so two rapid `remote-episode-change` events don't fight `goToEpisode`.

src/main/syncplay.ts
```diff
@@ -0,0 +1,724 @@
...
+    })
+    this.reconnectTimer = setTimeout(() => {
+      this.reconnectTimer = null
+      this.openSocket()
+    }, delay)
+  }
+
+  private setStatus(patch: Partial<SyncplayStatus>): void {
+    this.status = { ...this.status, ...patch }
+    this.emit('connection-status', { ...this.status })
```
akvasha, line 720: `setStatus` is a partial-merge, which means `error` is sticky across state transitions. After a `disconnected` event with an error, the next `openSocket()` setStatus (line 212) doesn't clear `error`, so the `connecting` / `reconnecting` event carries the prior error in its payload. `PlayerView.vue:1994` logs `status.error` whenever it's truthy, so the renderer console shows the stale error during every reconnect attempt.

Fix: have setStatus's spread include `error: undefined` by default, e.g. `this.status = { ...this.status, error: undefined, ...patch }`. Or have `openSocket` and `finishHandshake` explicitly pass `error: undefined`.

src/main/syncplay.ts
```diff
@@ -0,0 +1,724 @@
...
+    // five reconnect attempts on a misconfigured server. Disconnect cleanly
+    // with a clear error and stop. Detected by Node TLS error codes (all
+    // cert-trust failures start with `CERT_*` or `*_CERT_*`, plus the
+    // `ERR_TLS_*` family).
+    const code2 = typeof err.code === 'string' ? err.code : ''
+    const isFatalTls =
+      code2.startsWith('ERR_TLS_') ||
+      code2.includes('CERT_') ||
+      code2 === 'EPROTO' ||
+      (code2 === 'ECONNRESET' && this.status.state === 'tls-handshake')
```
akvasha, line 660: nit: `ECONNRESET` is treated as fatal-TLS only during `tls-handshake`. But the same reset during `tls-probing` (a server that doesn't speak Syncplay at all closes the TCP connection after our `{TLS:{option:'send'}}` line) lands here too, and gets classified as transient — so we burn 5 reconnect attempts at 1s, 2s, 4s, 8s, 16s before giving up. Add `tls-probing` to the fatal set.

```suggestion
      (code2 === 'ECONNRESET' && (this.status.state === 'tls-handshake' || this.status.state === 'tls-probing'))
```

src/main/syncplay.ts
```diff
@@ -0,0 +1,724 @@
...
+    if (!text) return
+    this.emit('room-event', { level: 'chat', text: user ? `${user}: ${text}` : text })
+  }
+
+  private handleError(payload: unknown): void {
+    if (!isObject(payload)) return
+    const text = typeof payload.message === 'string' ? payload.message : 'Server error'
+    console.warn('[syncplay] server error:', text)
+    this.emit('room-event', { level: 'error', text })
+    if (/version|password|banned/i.test(text)) {
```
akvasha, line 539: nit: `/version|password|banned/i.test(text)` is a brittle substring match against the human-readable server message. Real Syncplay servers send English text but a translated/customized server (or a future protocol bump) won't match. Worse: "version" matches "new feature: room-version sharing" or similar non-fatal info messages. The Syncplay protocol doesn't ship structured error codes, so this isn't easily fixable, but worth a TODO comment so it doesn't get forgotten.

PR #42, opened 2026-04-25
Fix auto-advance timer leak triggering multi-episode skip

### LINKED ISSUE #40: Player auto-advance loop triggers multiple episode skips
### Describe the bug
When reaching the end of an episode, the player displays a 5-second countdown timer before automatically advancing to the next episode. However, if the `ended` event on the `<video>` element is fired multiple times (e.g., due to MediaSource stream stuttering or the user clicking near the very end of the seekbar), it causes the player to rapidly skip through subsequent episodes all the way to the end of the season.

### Steps to reproduce
1. Open an episode in the built-in player.
2. Skip to the very end of the video (last 1-2 seconds).
3. Trigger the `ended` event multiple times (e.g. by clicking at the end of the seekbar repeatedly before the countdown finishes).
4. Wait for the countdown to reach zero.
5. The player navigates to the next episode. Once it loads, it will immediately skip to the next one again without waiting for the video to end.

### Root cause
There is a timer leak in `src/renderer/src/components/PlayerView.vue` inside the `onVideoEnded` function:
```javascript
function onVideoEnded(): void {
  if (!canNext.value) return
  autoAdvanceCountdown.value = 5
  autoAdvanceTimer = setInterval(() => { ... }, 1000)
}
```
If `onVideoEnded` is called a second time while the first countdown is still active, it creates a new `setInterval` and overwrites the `autoAdvanceTimer` variable. The first interval loses its reference and can never be cleared by `cancelAutoAdvance`. 
When the episode switches, the `navigating` lock prevents switching temporarily, but as soon as the next episode loads and `navigating` becomes false, the leaked timer (whose countdown is now `<= 0`) fires and instantly triggers another navigation, creating an infinite loop until it runs out of episodes.

akvasha:
## Summary
- Guard `onVideoEnded` against re-entry while a countdown is already active in `src/renderer/src/components/PlayerView.vue`.
- Prevents stacked `setInterval` timers from leaking and triggering rapid sequential episode skips.

Fixes #40

## Root cause
Multiple `ended` events (e.g. from MediaSource stutter or seekbar clicks near the end) each created a new `setInterval` and overwrote `autoAdvanceTimer`, leaving prior intervals unreferenced. A leaked timer's countdown reaches zero while `navigating` is locked, then fires immediately once the next episode loads — cascading through the season.

## Test plan
- [x] Open episode in built-in player, seek to last 1–2s, click seekbar end repeatedly, confirm only one countdown overlay appears and only one episode advance happens.
- [x] Cancel button still aborts the countdown.
- [x] Normal end-of-episode auto-advance still works.

### DIFF
--- a/src/renderer/src/components/PlayerView.vue
+++ b/src/renderer/src/components/PlayerView.vue
@@ -1575,6 +1575,7 @@ function cancelAutoAdvance(): void {
 
 function onVideoEnded(): void {
   if (!canNext.value) return
+  if (autoAdvanceTimer) return
   autoAdvanceCountdown.value = 5
   autoAdvanceTimer = setInterval(() => {
     autoAdvanceCountdown.value--


mikkerlo (approved):
This PR introduces a crucial bug fix that resolves an unexpected edge-case in our built-in video player where multiple `ended` events triggered an uncontrollable auto-advance loop through the remaining episodes.

### General Feedback
Great job identifying the root cause and applying the simplest, most effective fix. This prevents a very jarring user experience where the user would lose control of their viewing session after a simple seek or playback stutter near the end of an episode.

### Specific Observations
- **Surgical Fix:** Adding the `if (autoAdvanceTimer) return` guard is the perfect low-impact solution. It ensures that the first triggered countdown completes its cycle while aggressively dropping any spurious redundant `ended` events.
- **Clean Architecture:** It perfectly utilizes the existing state (`autoAdvanceTimer`) rather than introducing complex new flags or state machines.
- **Immediate Impact:** This directly addresses Issue #40 without risking regressions in other parts of the playback lifecycle.

- **Future Proofing:** While the fix is perfectly adequate for the current implementation, if the auto-advance logic becomes more complex in the future, we might consider combining this with a check on `navigating.value` to ensure we don't accidentally start a timer while already mid-transition. However, the current fix covers the core leak beautifully.

LGTM after manual testing

PR #45, opened 2026-04-26
test: bot review-collapse

akvasha:
Throwaway PR to test the Telegram bot's review-collapse behavior. Will be closed immediately after testing.

PR #44, opened 2026-04-26
ci: skip PR artifact build when only docs changed

akvasha:
## Summary
PR #43 (a TODO-only docs change) triggered the full 3-platform artifact build, which is wasted CI for changes that touch no shipping code. This PR adds a `changes` job using `dorny/paths-filter` to detect whether a PR touches anything other than docs, and gates the `build` matrix on the result.

- Docs-only PRs (only `*.md`, `LICENSE`, `.gitignore`, `.gitattributes`) → `build` and `comment-artifacts` skip cleanly. Both report as "skipped" rather than failed, so required-status-check semantics are preserved.
- Code PRs → unchanged behavior: full build + artifacts comment.
- `/build` issue-comment override → still forces a build regardless of touched paths (so reviewers can manually request artifacts on a docs PR if needed).

## Test plan
- [ ] Open a docs-only PR after merge; confirm the build matrix is skipped and no artifacts comment is posted.
- [ ] Open a code PR after merge; confirm the build matrix runs as before and the artifacts comment posts.
- [ ] Comment `/build` on a docs PR; confirm the override still triggers a build.

PR #46, opened 2026-04-26
ci: rewrite paths-filter as allowlist of build-relevant paths

akvasha:
## Summary

PR #44 tried to skip the multi-platform artifact build on docs-only PRs but didn't actually work. PR #45 (a markdown-only test PR) still triggered all three builds. The workflow run showed `dorny/paths-filter` evaluating `code = true` and listing `TEST_BOT_REVIEW.md` as a matching file.

## Root cause

The previous denylist relied on negated patterns under `paths-filter`'s default `predicate-quantifier: some` (OR). With `some`, a file matches the filter if **any** rule matches:

- `**` always matches every file → `true`
- `!**/*.md` is a separate inverse rule → `false` only for `.md` files

Since `**` is always `true`, the negations never excluded anything.

## Fix

Switch to an explicit allowlist of paths that feed into `electron-vite` + `electron-builder`:

- `src/**`
- `resources/**`
- `package.json`
- `package-lock.json`
- `electron.vite.config.*`
- `tsconfig*.json`

Anything outside this list — `.github/` workflow edits, `*.md` docs, `LICENSE`, `scripts/`, root configs — is now a no-op for the build. The `/build` comment escape hatch still bypasses the filter when needed.

## Test plan

- [x] Reproduced bug on PR #45 (markdown-only branch built all three platforms).
- [ ] This PR is itself a `.github/` change → with the new filter, the `build` jobs should be skipped (only `changes` runs).
- [ ] Follow-up scratch PR touching `src/main/index.ts` only should still trigger the build.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -29,14 +29,18 @@ jobs:
         id: filter
         if: github.event_name == 'pull_request'
         with:
+          # Allowlist: build only when files that actually feed into the
+          # electron-vite + electron-builder output change. Anything outside
+          # this list (docs, .github/, LICENSE, scripts unrelated to the bundle)
+          # is treated as a no-op and skips the artifact build.
           filters: |
             code:
-              - '**'
-              - '!**/*.md'
-              - '!*.md'
-              - '!LICENSE'
-              - '!.gitignore'
-              - '!.gitattributes'
+              - 'src/**'
+              - 'resources/**'
+              - 'package.json'
+              - 'package-lock.json'
+              - 'electron.vite.config.*'
+              - 'tsconfig*.json'
 
   build:
     needs: changes


mikkerlo (approved):
This PR perfectly addresses the underlying issue with the previous attempt to skip docs-only builds. Switching from a brittle denylist to an explicit allowlist is absolutely the right architectural move for CI triggers.

### General Feedback
Excellent debugging! You accurately pinpointed how `dorny/paths-filter` evaluates multiple patterns with `predicate-quantifier: some`. An allowlist is inherently safer because it fails closed (skips the heavy build) for any newly added repository administration files.

### Specific Observations
- **Root Cause Analysis:** Spot on regarding the `**` pattern overpowering the negated `!` rules. This is a very common gotcha with glob evaluation logic in actions.
- **Precision Targeting:** The targeted allowlist (`src/**`, `resources/**`, `package.json`, `electron.vite.config.*`, `tsconfig*.json`) perfectly encapsulates the actual dependency graph for the `electron-vite` bundle and the `electron-builder` packaging step. 

- Everything looks solid. The `/build` override remains functional, meaning we maintain full escape-hatch capability.

LGTM after manual testing

PR #47, opened 2026-04-26
feat: continue watching home view (v3.13.0)

akvasha:
## Summary

- New default landing tab that surfaces **what to play next** — a lifted version of TODO #5.
- **Resume** rows from local `watchProgress` (collapsed to the most-recent unfinished episode per anime, so a show with progress on episodes 5/6/7 appears as a single card).
- **Next-up** rows from Shikimori `watching`/`rewatching` rates where the next episode has aired.
- Sort is the **same rule as the Shikimori "To Watch" tab** — descending `rate.updated_at`. Resume rows for tracked anime inherit the Shikimori clock so the order stays consistent across views; resume rows for anime not on Shikimori fall back to local `watchProgress.updatedAt`.
- Click a card → `AnimeDetailView` opens focused on the targeted episode (handles pagination + `scrollIntoView`).
- New `recentAnimeMeta` store + lazy `smotretApi.getAnime` fallback so streamed-only anime resolve titles/posters instead of showing as "Unknown anime" with a black placeholder.

## Test plan

- [ ] Home loads on first launch (instead of Search).
- [ ] Resume cards appear for anime with partial progress; clicking jumps to the right page (paginated anime) and scrolls the row into view.
- [ ] An anime with progress on multiple episodes appears as a single Resume card pointing at the most-recent episode.
- [ ] Next-up cards appear for Shikimori watching/rewatching where `rate.episodes + 1 <= episodes_aired`.
- [ ] Anime that exists in both buckets emits only the Resume card (dedup).
- [ ] Order matches the Shikimori "To Watch" tab.
- [ ] Streamed-only anime (no library/downloads entry) eventually resolve their title + poster after refresh.
- [ ] Empty profile shows the "Browse Search" CTA.
- [ ] Sidebar Home/Search/Library/etc. all navigate; back from AnimeDetailView returns to Home.


mikkerlo (approved):
This PR is a fantastic addition, effectively fulfilling the "Continue Watching Home View" TODO item by introducing a polished new default landing page. It elegantly merges local playback progress and Shikimori user rates into a cohesive, priority-sorted feed.

### General Feedback
The feature is implemented beautifully, from the unified data sorting down to the smooth UX of jumping straight into a paginated list. The data fetching approach in `home:get-continue-watching` is robust, performing local merges quickly while safely lazy-loading missing metadata.

### Specific Observations
- **Deep Linking:** The `applyFocusEpisode` logic in `AnimeDetailView.vue` is a massive UX win. It seamlessly handles pagination offsets and uses `scrollIntoView` correctly without causing layout thrash.
- **Lazy Resolution:** Introducing `recentAnimeMeta` gracefully solves the edge-case for streamed-only anime (that have no downloaded or library record) without a bloated cache footprint. The 5-second `Promise.race` timeout on background metadata resolution is a solid defensive choice.
- **Smart Deduplication:** The logic properly collapses multiple `watchProgress` episodes for a single show into just the most recently updated unfinished one, and correctly prioritizes `Resume` rows over `Next` rows.
- **Sorting Integration:** Overriding the sort key of Resume rows with Shikimori's `updated_at` clock ensures flawless chronological continuity across the local and remote ecosystems.

- **Pre-existing Technical Debt Note:** In `HomeView.vue` (and previously in `AnimeDetailView` and `ShikimoriView`), the `onUnmounted` hook calls `window.api.offShikimoriRateUpdated()` and similar methods. Because the current IPC preload implementation for these `off*` methods uses `ipcRenderer.removeAllListeners()`, unmounting ANY component that calls these cleanup methods strips the listeners for ALL currently mounted components. Since `HomeView` is bound with `v-show`, it effectively never unmounts during normal use (so it won't actively trigger this bug), but the same bug exists when `AnimeDetailView` unmounts. This is worth tracking for a future IPC refactor to use explicit listener ID mapping to safely remove only specific listeners, but it is outside the scope of this PR.

LGTM after manual testing

mikkerlo:
Bug: HomeView shows "Трейлер" instead of episode 1 for `Guimi Zhi Zhu: Xiaochou Pian` (smotret id 34496).

**Root cause:** the smotret-anime API returns this anime's episodes with a trailer entry that shares `episodeInt` with the real episode:

```
episodeInt='1', episodeType='preview', episodeFull='Трейлер'
episodeInt='1', episodeType='ona',     episodeFull='ONA 1 серия'
episodeInt='2', episodeType='ona',     episodeFull='ONA 2 серия'
...
```

The new `home:get-continue-watching` handler resolves labels with `Array.find` keyed only on `episodeInt`, so for any anime whose API listing leads with a `preview`/trailer entry the trailer wins:

```ts
// src/main/index.ts (added by this PR)
function episodeLabelFor(animeId: number, episodeInt: string): string {
  const entry = cache[String(animeId)]
  const ep = entry?.animeDetail?.episodes?.find((e) => e.episodeInt === episodeInt)
  return ep?.episodeFull || `Episode ${episodeInt}`
}
```

`AnimeDetailView.vue` already excludes these via `ep.episodeType !== 'preview'` when building its episode list, so the detail view and on-disk filenames (`... - 01 [Author].mp4`) are correct — only the new HomeView label is wrong.

**Patch:** apply the same `episodeType !== 'preview'` exclusion in both `find` callbacks added by this PR.

```ts
// episodeLabelFor (~line 140 of the diff)
const ep = entry?.animeDetail?.episodes?.find(
  (e) => e.episodeInt === episodeInt && e.episodeType !== 'preview'
)

// Next-up smotret fallback (~line 330 of the diff)
if (r.episodeLabel.startsWith('Episode ')) {
  const ep = meta.episodes?.find(
    (e) => e.episodeInt === r.episodeInt && e.episodeType !== 'preview'
  )
  if (ep?.episodeFull) r.episodeLabel = ep.episodeFull
}
```

Optionally factor the predicate into a shared helper (e.g. `isContentEpisode`) so any future call site stays consistent with `AnimeDetailView`'s filter.

akvasha:
1. **Trailer label bug (Guimi Zhi Zhu: Xiaochou Pian)** — fixed. Both `Array.find` call sites in `home:get-continue-watching` (`episodeLabelFor` for cached entries, plus the lazy-fetch apply loop for newly-resolved entries) now match on `episodeInt && episodeType !== 'preview'`, mirroring AnimeDetailView's existing filter. Factored the predicate into a small `isContentEpisode` helper so any future call site stays consistent.

Pushed as a follow-up commit on top of the PR.

mikkerlo:
Fixed!

PR #58, opened 2026-04-30
feat: storage usage dashboard and auto-cleanup of watched episodes

### LINKED ISSUE #54: Storage Dashboard & Auto-Cleanup of Watched Episodes

**Motivation:** Downloaded anime accumulates indefinitely. Users have no way to see *what's eating disk space* or to automatically free space after they've finished watching. A Storage tab in Settings answers "what's on disk and where?" and an opt-in cleanup rule ("delete watched episodes after 30 days") closes the loop without manual housekeeping.

**Plan:**
1. **Schema additions to `watchProgress`.**
   - Add an explicit `watchedAt?: number` field (epoch ms) recorded the first time `watched` flips false→true. Distinct from `updatedAt`, which keeps mutating during heartbeats.
   - In `src/main/index.ts` `watch-progress:save` handler: when `watched === true` AND `prev?.watched !== true`, set `watchedAt = Date.now()`. Otherwise carry `prev?.watchedAt` forward unchanged. (Don't overwrite on rewatches — first-watched timestamp is the policy anchor.)
   - Update the store type definition + `WatchProgressEntry` in `src/preload/types.d.ts`.
2. **Main: new IPC `storage:get-usage`.**
   - Scans hot dir + cold dir (when advanced mode) and returns:
     ```ts
     { totalBytes: number, perAnime: Array<{ animeId: number, animeName: string, bytes: number, fileCount: number, episodes: Array<{ episodeInt: string, files: { mkv?: { path, size }, mp4?: { path, size }, ass?: { path, size } }, watched: boolean, watchedAt?: number }> }> }
     ```
   - Implementation: `fs.readdirSync` each anime folder, group files by sanitized anime name → match against `downloadedAnime` keys, group within an anime by parsing episode number from filename (regex matches the `{name} - {NN} [Author].ext` shape used in `download-manager.ts`), sum sizes via `fs.statSync(path).size`. Cross-reference each `(animeId, episodeInt)` against `watchProgress` for the watched flag + `watchedAt`.
   - Sorted: anime by `bytes` desc; episodes by `episodeInt` asc.
   - Long-running on big libraries — async with `fs.promises`. Fire `storage:usage-progress` IPC events with `{ scanned, total }` if the library has >50 anime, so the UI can show a progress bar. Result returned via the regular handle when complete.
3. **Main: new auto-cleanup worker.**
   - Settings: `autoCleanupWatchedDays: number` (0 = off, default 0). `autoCleanupConfirm: boolean` (default `true` — first run shows a confirm modal listing what will be deleted).
   - Worker `runWatchedCleanup(force = false): Promise<{ deletedCount, freedBytes, items }>`:
     a. Read `autoCleanupWatchedDays`. If 0 and not `force`, return early.
     b. Walk `watchProgress` for entries with `watched === true` AND `(now - watchedAt) > days * 86400_000`.
     c. For each, parse `${animeId}:${episodeInt}` and call the existing `file:delete-episode` flow (or a shared internal helper) to remove all translations of that episode (from hot + cold). Sum freed bytes from the pre-delete file sizes.
     d. Emit a `storage:cleanup-finished` IPC event with the result; broadcast `file:episodes-changed` for any anime that lost files.
   - Trigger cadence: on app start (after a 60 s warmup), and once per 24 hours via `setInterval`. Plus an on-demand `storage:run-cleanup` IPC for the "Run now" button.
   - First-time gate: if `autoCleanupConfirm === true` AND a scheduled run would delete >0 episodes, instead emit `storage:cleanup-pending` with the candidate list; the renderer shows a modal; on confirm, the renderer calls `storage:run-cleanup` with `force: true` and then `setSetting('autoCleanupConfirm', false)`. Subsequent runs are silent.
4. **Preload (4-file pattern).**
   - Expose: `storageGetUsage`, `storageRunCleanup`, `onStorageUsageProgress`, `onStorageCleanupFinished`, `onStorageCleanupPending`. Add types.
5. **Renderer: new "Storage" tab in `SettingsView.vue`.**
   - Top: total bytes used (large number), counts `(N anime · M episodes · K files)`, refresh button.
   - Anime list: each row collapsible. Header shows poster thumb, title, total bytes, file count, expand chevron. Body lists episodes with size, watched checkmark + relative `watchedAt`, and a per-episode delete button (uses existing `file:delete-episode`).
   - Below the list: "Auto-cleanup" panel. Number input for `autoCleanupWatchedDays` (0–365; 0 = off). Helper text: "Delete watched episodes older than N days". Status line: last run, episodes deleted, bytes freed (persist in store as `autoCleanupLastRun`). "Run cleanup now" button → `storageRunCleanup` (still respects the confirm modal on first run).
   - Confirm modal (shown on first scheduled or manual run when `autoCleanupConfirm === true`): lists candidate anime + episode counts; buttons `Delete` / `Cancel`. On Delete, persists `autoCleanupConfirm = false` + invokes `storageRunCleanup({ force: true })`.
6. **Migration for existing `watchProgress` entries without `watchedAt`.**
   - On boot, one-time pass: for each entry with `watched === true` AND `watchedAt == null`, set `watchedAt = updatedAt`. Mark migration done via `watchProgressMigrationV2: true` setting so we don't re-run.
7. **DESIGN.md update.** New "Storage Dashboard & Auto-Cleanup" section. Document: the `watchedAt` field, the cleanup worker cadence + confirm-modal flow, the `storage:get-usage` shape, the migration. Add new IPC channels to the table and new settings to the Settings table.

**Blockers & Risks:**
- **Cleanup trust.** Deleting files automatically is the riskiest action this app does. The default is **off**, the first run is gated on a confirm modal, and even after enabling, every deletion goes through the same `file:delete-episode` path that the user already uses manually — no parallel deletion code path. Mitigation: also log every deletion to a `cleanupLog` setting (rolling, last 100 entries) that the Storage tab can show as "Recent cleanups".
- **`watchedAt` accuracy on legacy data.** Pre-migration entries get `watchedAt = updatedAt`, which is roughly correct (the last save when watched was set), but for entries where the user re-opened the episode after marking it watched, `updatedAt` could be much later than when they actually finished. Acceptable — errs on the side of *not* deleting.
- **In-progress downloads.** Cleanup must skip episodes with active downloads or `.part` files. The existing `moveEpisodeToColdStorage` already skips `.part`-suffixed files; reuse the same guard.
- **Multi-translation episodes.** When watched is set on a `(animeId, episodeInt)`, all translations of that episode get deleted together by the existing `file:delete-episode` (without `translationId`). That's the right behavior — once the user is done with that episode, every author's version of it is fair game. Worth calling out in the Settings helper text.
- **Scan performance.** `storage:get-usage` on a library of 100+ anime with 1000+ files takes seconds, not ms. Justifies the progress IPC + async impl. Don't block the Settings tab open on this — render the tab immediately, populate when the scan resolves.
- **Cold storage interaction.** On systems with hot/cold split, `storage:get-usage` reports both dirs (separately broken out, e.g. `bytesHot` / `bytesCold`). Cleanup deletes from whichever location holds the file (existing `file:delete-episode` already handles both).

akvasha:
Closes #54.

Implements the Storage Dashboard & Auto-Cleanup proposal in full: disk-usage scanner, opt-in cleanup-after-watched policy with a first-run confirm gate, plus the `watchProgress` schema bump and migration. UI lives in a new section on the existing Settings → Storage tab (no separate tab — it shares space with hot/cold config since they're related concerns).

## Scope vs. issue plan

- ✅ `watchedAt` schema bump and one-time `watchProgressMigrationV2` (item 1).
- ✅ `storage:get-usage` with hot/cold split, per-anime/per-episode breakdown, file map by extension, and `storage:usage-progress` events for libraries >50 anime (item 2).
- ✅ `runWatchedCleanup` worker: warmup at 60s, then every 24h, plus on-demand `storage:run-cleanup`. Honors `autoCleanupConfirm` first-run gate via `storage:cleanup-pending` → renderer modal → `force: true` callback (item 3).
- ✅ Preload bindings + types (item 4).
- ✅ Renderer dashboard with poster/title rows, expandable episode lists, format tags, watched indicator, per-episode delete, auto-cleanup days input, "Run now" button, last-run status, and rolling 100-entry history log (items 5, 6).
- ✅ Reuses the existing `file:delete-episode` deletion path (refactored into a shared `deleteEpisodeFiles` helper) so manual and automated paths stay in lockstep — addresses the "cleanup trust" risk.
- ✅ Skips episodes with in-progress `.part` files via `episodeHasInProgressDownload` — addresses the in-progress-download risk.
- ⏭️ DESIGN.md update (item 7) deferred to a follow-up — happy to add in this PR if preferred.

## Folder-name fix

Surfaced during testing: the scanner originally keyed folders by `sanitizeFilename(anime.title)`, but downloads are written under `sanitizeFilename(getAnimeName(anime))`, which prefers `titles.romaji → titles.ru → title`. Result: scans showed 0 B for most libraries because folders were named after the romaji/ru title. The lookup map now registers all three candidates and surfaces the preferred name (matching the on-disk folder) so subsequent delete calls round-trip cleanly. Same fix applied to `findCleanupCandidates`.

## Test plan

- [ ] Open Settings → Storage, click **Scan storage**: verify totals, hot/cold breakdown (advanced mode), per-anime expansion, episode tags, and per-episode delete.
- [ ] Mark some episodes watched, set auto-cleanup days low, click **Run cleanup now**: verify confirm modal lists candidates; confirm and check files are deleted, history populates, and storage rescan reflects freed space.
- [ ] Confirm subsequent cleanup runs (after the first confirm) skip the modal.
- [ ] Verify nothing is deleted while a `.part` download is in progress for the same episode.
- [ ] Verify advanced-mode cold-storage entries take priority over hot duplicates in usage totals.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "anime-dl-app",
-  "version": "3.13.0",
+  "version": "3.14.0",
   "description": "Anime episode downloader",
   "main": "./out/main/index.js",
   "scripts": {
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -83,7 +83,12 @@ const store = new Store({
     playerMuted: false,
     anime4kPreset: 'off' as 'off' | 'mode-a' | 'mode-b' | 'mode-c',
     hevcTranscodeOnPlay: 'ask' as 'ask' | 'always' | 'never',
-    watchProgress: {} as Record<string, { position: number; duration: number; updatedAt: number; watched?: boolean }>,
+    watchProgress: {} as Record<string, { position: number; duration: number; updatedAt: number; watched?: boolean; watchedAt?: number }>,
+    watchProgressMigrationV2: false,
+    autoCleanupWatchedDays: 0,
+    autoCleanupConfirm: true,
+    autoCleanupLastRun: null as { ranAt: number; deletedCount: number; freedBytes: number } | null,
+    cleanupLog: [] as { ranAt: number; animeId: number; animeName: string; episodeInt: string; bytes: number }[],
     shikimoriUserRates: [] as unknown[],
     shikimoriUpdateQueue: [] as unknown[],
     shikimoriAnimeDetails: {} as Record<string, { details: shikimori.ShikiAnimeDetails; fetchedAt: number }>,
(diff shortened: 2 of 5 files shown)


mikkerlo (approved):
### General Feedback
The storage dashboard and auto-cleanup feature looks great and correctly implements the planned functionality. The split between hot/cold storage usage calculation is well done, and `deleteEpisodeFiles` handles the shared deletion path effectively. Good use of the existing IPC patterns to propagate state to the renderer.

### Specific Observations
- The fallback logic for folder name lookup across `titles.romaji`, `titles.ru`, and `title` correctly fixes the 0B size issue and ensures backward compatibility with older folders.
- `episodeHasInProgressDownload` correctly bails out when `.part` files are detected, nicely mitigating the risk of deleting active downloads.
- The `cleanupPending` UI works cleanly as a modal atop the settings page.

LGTM after manual testing

Issue #59, opened 2026-05-01
Find OP/ED button: local Chromaprint fingerprinting for skip detection


akvasha:
**Priority:** Medium | **Effort:** Large

Replace aniskip's crowdsourced timestamps with per-encode-accurate OP/ED detection by fingerprinting the user's actual downloaded episodes via Chromaprint (`fpcalc`). Aniskip stays as a cold-start fallback only. Discussed on PR #55: https://github.com/mikkerlo/anime-downloader/pull/55#issuecomment-4361913726

### Why

Aniskip (shipped in v3.14.0 / PR #55) uses timestamps tied to a reference encode of each show. In practice they often don't match smotret-anime.ru's encode — observed on *Frieren* (intro starts ~6s before OP, skip target lands ~6s after OP end). Fingerprinting the user's actual file is correct *for that encode* by construction.

### Approach

Same algorithm Jellyfin's [intro-skipper](https://github.com/intro-skipper/intro-skipper) uses: for each show, compare ≥2 downloaded episodes via Chromaprint hash streams; the long contiguous low-Hamming-distance run between them is the OP (second match = ED). No external database needed; comparison happens locally between episodes of the same season.

### Plan

1. **Bundle `fpcalc`** — new module mirroring `ffbinaries` pattern. On first launch, download platform binary from [acoustid/chromaprint releases](https://github.com/acoustid/chromaprint) (~3 MB) into the userData dir. Store path alongside ffmpeg path.

2. **`src/main/fingerprint.ts`** — wrapper around `fpcalc`. Single export:
   ```ts
   export async function probe(source: string, durationSec: number): Promise<number[]>
   ```
   Pipes ffmpeg → fpcalc:
   ```
   ffmpeg -i <source> -t <duration> -vn -f wav - | fpcalc -raw -
   ```
   Returns the raw 32-bit hash array. Same code path for local files and remote URLs (ffmpeg handles both transparently — relevant for the future server/online-probe phase but not this issue).

3. **`src/main/skip-detector.ts`** — replaces `src/main/aniskip.ts` as the primary IPC entry point (aniskip becomes an internal cold-start fallback). Owns:
   - `skipSignatureCache: Record<animeId, { opSignature, edSignature, derivedFromEpisodes[], confirmedAt }>`
   - `skipProbeCache: Record<animeId:episodeInt:translationId, { fingerprint, fetchedAt }>`

[...]

PR #60, opened 2026-05-02
feat: local Chromaprint OP/ED detection — Phase 1 (debug panel) + Phase 2 (player skip)

akvasha:
## Summary

Phases 1 + 2 of #59. Replaces (eventually) the planned crowdsourced AniSkip integration with per-encode-accurate OP/ED detection by fingerprinting the user's actual downloaded episodes via Chromaprint (`fpcalc`).

Rollout plan in full lives in `DESIGN.md` → "Skip Detection" → "Rollout phases". This PR ships **Phase 1** (local pipeline + debug panel) and **Phase 2** (player integration).

### Phase 1 — local pipeline + debug panel

- Bundles `fpcalc` (Chromaprint v1.5.1) auto-download mirroring the `ffbinaries` pattern (~5 MB tar.gz/zip from `acoustid/chromaprint` releases).
- New main-process modules:
  - `src/main/fpcalc-binaries.ts` — first-launch download + extract via system `tar`.
  - `src/main/fingerprint.ts` — spawns `fpcalc -raw -length 0 <file>`. Note: feeding raw PCM via stdin yields `DURATION=0` because there's no container metadata; fpcalc's bundled FFmpeg handles decoding when given a path.
  - `src/main/skip-detector.ts` — pairwise Hamming-distance sliding-window match over `popcount32` of XOR'd hashes; coarse boundaries are tightened by an *inward-only* per-hash refinement pass; per-episode OP/ED times aggregated by median across pairs.
- Persistent caches in electron-store: `skipDetections` per anime, `skipFingerprintCache` keyed by `animeId:episodeInt:fileSize:mtime` (so re-encoded files re-fingerprint automatically).
- IPC: `skip-detector:analyze-show`, `:get-detections`, `:get-status` (re-mount recovery), `:cancel`, `:cache-stats`, `:analyze-progress` (broadcast).
- AnimeDetailView gets a collapsible `Skip Detection (debug)` panel: states for disabled (<2 episodes), idle, analyzing+cancel, and a results table with OP/ED `mm:ss–mm:ss` per episode plus pair counts.

### Phase 2 — player integration

- `PlayerView` calls `skipDetectorGetDetections(animeId)` on mount.
- **Seek-bar bands**: green OP / blue ED tints inside `.seek-track`, positioned by start/end percentages of duration. `pointer-events: none` so they don't block the seek input on top. The existing seek tooltip is extended to append ` · OP` / ` · ED` when the hover position falls inside a band (the title attribute never fires under `pointer-events: none`).
- **Skip OP/ED button overlay**: anchored bottom-right. `activeSkipRange` computes from `currentTime` (with a 0.25 s lead-in tolerance), returns `'op' | 'ed' | null`. A 250 ms grace timer prevents flicker when scrubbing. Click → `seek(range.endSec)` + adds the `(animeId, episodeInt, kind)` key to `skippedRanges` so the button does not reappear on rewind. The set resets on episode change.
- No new IPC needed.

### Algorithm notes

- Coarse: 6 s window, 18 s min run, average distance threshold 6/32 bits.
- Refinement: 5 consecutive hashes ≤ 4/32 bits, scanned inward from each coarse edge — never extends outward (the original outward refinement pushed boundaries earlier rather than tightening).
- Verified on Re:Zero S4 ep 2: detected OP within ~2 s of visually observed onset.

### Phase 3 / 4 (separate follow-up PRs)

- **Phase 3**: auto-trigger background analysis on `onEpisodeComplete` + `onMergeComplete` once ≥ 2 episodes are on disk; `skip-detector:signature-updated` broadcast for live UI updates without polling; `enableLocalSkipDetection` user setting.
- **Phase 4**: GC `skipFingerprintCache` on file delete; demote AniSkip to cold-start fallback if/when it lands.

## Test plan

- [x] `npm run build` succeeds
- [x] First launch downloads fpcalc to userData/fpcalc and chmods 0o755
- [x] AnimeDetailView debug panel: Analyze + Re-analyze + Cancel work; cancel kills child fpcalc processes; re-analysis hits cache and is significantly faster
- [x] PlayerView: green OP / blue ED bands appear on the seek bar for analyzed shows; tooltip shows ` · OP` / ` · ED` inside bands
- [x] Skip OP/ED button appears with 250 ms grace, jumps to range end, doesn't re-appear on rewind, resets on episode change
- [ ] Verify on Windows + macOS (system `tar` extraction path) — needs another machine


akvasha:
Thanks for the review! Mapping each point to phases:

**1. Auto-trigger on `onEpisodeComplete` AND `onMergeComplete`** — agreed. Deferred to Phase 2 since auto-trigger is explicitly out of scope for this PR (manual trigger only via the debug panel). When it lands, we'll hook both callbacks so the analysis runs regardless of `autoMerge`.

**2. `skip-detector:signature-updated` broadcast** — also Phase 2. Phase 1 only runs analysis from a UI click, so the await-ing renderer always gets the result directly. Once auto-trigger lands and analysis can complete with no UI listening, we'll add the broadcast (mirroring the Shikimori `rate-updated` pattern).

**3. UI re-mount recovery** — fixed in [01fc7cd](https://github.com/mikkerlo/anime-downloader/commit/01fc7cda) on top of the original commit:
- Replaced the single `AbortController` with a module-level `currentAnalysis` object (`{ animeId, controller, lastProgress, promise }`).
- `analyze-show` for the *same* `animeId` now dedupes onto the in-flight promise (so panel re-mounts or duplicate clicks ride the existing run instead of starting a new one).
- `analyze-show` for a *different* `animeId` rejects with a clear error instead of silently killing the original analysis — addresses the surprising-cancellation case directly.
- New `skip-detector:get-status` IPC returns `{ animeId, lastProgress }` for the active analysis (or null). `AnimeDetailView` calls it on mount to hydrate `skipAnalyzing` + `skipProgress` when re-opening the panel during a long run.
- Progress listener now treats `phase === 'done'` as the completion signal and reloads detections — matters when the analysis finishes after the panel was unmounted and re-mounted.

**4. Storage schema** — defaults are wired up under slightly different names (`skipDetections` and `skipFingerprintCache`); the cache key includes `fileSize:mtime` so re-encoded/replaced files automatically re-fingerprint, but pure deletion-driven GC isn't there yet — stale entries don't cause incorrect results (the algorithm only looks up keys for currently-present files), just unbounded growth. Worth tackling alongside the file-delete-cache pattern in `download:cancel-by-episode` / `file:delete-episode` as a small follow-up — happy to land that here if you'd prefer, or split into another PR.

Let me know if 1, 2, or the GC for 4 should land in this PR after all.

mikkerlo:
From my analysis works wonderful for at least 3 episodes downloaded. This is ready to be in release, let's mark menu as experimental though.

src/renderer/src/components/PlayerView.vue
```diff
@@ -141,6 +141,107 @@ const playerShortcuts = ref<Record<string, string>>({ ...DEFAULT_PLAYER_SHORTCUT
...
+
+// Skip Detection (Phase 2): show OP/ED bands on the seek bar and a "Skip OP/ED"
+// overlay button when playback enters a detected range. Boundaries come from
+// `skipDetectorGetDetections(animeId)` — no new IPC needed (Phase 1 already
+// persists per-episode boundaries in electron-store).
+const showSkipDetections = ref<ShowSkipDetections | null>(null)
+const skippedRanges = ref<Set<string>>(new Set())
+const skipButtonVisible = ref(false)
+let skipButtonGraceTimer: ReturnType<typeof setTimeout> | null = null
+const SKIP_GRACE_MS = 250
```
mikkerlo, line 153: nit: `skipButtonGraceTimer` is created here but isn't cleared in `onBeforeUnmount`. If a user navigates away from `PlayerView` while the 250ms grace timer is running, the callback will execute on an unmounted component, potentially causing a memory leak or error.

It would be good to add `if (skipButtonGraceTimer) clearTimeout(skipButtonGraceTimer)` to the existing `onBeforeUnmount` block.

src/renderer/src/components/AnimeDetailView.vue
```diff
@@ -1334,6 +1453,89 @@ function typeChip(type: string): { short: string; color: string } {
...
+        <div class="skip-header" @click="skipPanelCollapsed = !skipPanelCollapsed">
+          <div class="skip-header-left">
+            <svg
+              class="skip-chevron"
+              :class="{ collapsed: skipPanelCollapsed }"
+              viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" width="14" height="14"
+            >
+              <path stroke-linecap="round" stroke-linejoin="round" d="M6 9l6 6 6-6"/>
+            </svg>
+            <span class="skip-label">Skip Detection (debug)</span>
```
mikkerlo: Per mikkerlo's comment, let's mark this as experimental rather than just debug.

```suggestion
            <span class="skip-label">Skip Detection (experimental)</span>
```

mikkerlo (approved):
Great work on Phase 1 and 2! The architecture aligns perfectly with `DESIGN.md` and the extraction/fingerprinting logic looks solid.

Per mikkerlo's recent comment, I've left a suggestion to mark the panel as experimental rather than debug. I also caught one minor missing timer cleanup in `PlayerView.vue`.

LGTM after manual testing.

akvasha:
Thanks for the review! Both addressed in 2804bca4:

1. **`PlayerView.vue` — `skipButtonGraceTimer` cleanup**: Fixed. Added a `clearTimeout(skipButtonGraceTimer)` block to `onBeforeUnmount` next to the other timer cleanups so the 250ms grace callback can't fire on an unmounted component.
2. **`AnimeDetailView.vue` — label wording**: Fixed. Renamed the panel header from "Skip Detection (debug)" to "Skip Detection (experimental)".

Typecheck clean.

PR #62, opened 2026-05-02
feat: Phase 4 — skip-detector cache GC

akvasha:
## Summary

Phase 4 of issue #59. The Phase 1 fingerprint cache is keyed by `animeId:episodeInt:fileSize:mtime` and grew unbounded — every successful merge produces a new key (mkv vs the source mp4 it was built from), every re-download produces a new key (mtime differs), and deletes orphan their entries. After a few weeks of normal use the cache pages start to get… big. This PR plugs the leak.

## What changes

- **Event-driven GC** — `file:delete-episode` and `downloaded-anime-delete` IPC handlers now call `pruneSkipCacheForEpisode(animeId, episodeInt)` and `pruneSkipCacheForAnime(animeId)` respectively. They also drop the corresponding `skipDetections` entries so the player no longer surfaces OP/ED for an episode the user just removed.
- **Startup sweep** — `sweepSkipFingerprintCache()` runs ~30 s after launch as a backstop. Per anime in the cache: scan the folder, stat each file, drop any cache entry whose `(fileSize, mtime)` doesn't match a live file (or whose anime isn't in `downloadedAnime` anymore). Catches external `rm`, in-place rename, file-replaced, and the very common case of a leftover `.mp4` cache entry after the merge step unlinked the source.
- DESIGN.md updated; version bumped 3.17 → 3.18.

## Out of scope

- **AniSkip cold-start fallback.** Phase 4's other half in DESIGN.md is conditional (*"if/when an AniSkip module exists"*) and there's no AniSkip module in the repo yet. When one lands, the natural pattern is: serve crowdsourced AniSkip times from `skip-detector:get-detections` for shows where local analysis hasn't run yet, then replace them once a local signature is computed. Tracked memory says AniSkip is provisional and should stay encapsulated behind a single `src/main/aniskip.ts` + IPC surface — a future PR can drop that in without touching this PR's GC.

## Test plan

- [ ] Delete a single downloaded episode → confirm the matching cache entry is gone (check `skip-detector:cache-stats`) and the per-episode entry in `skipDetections` for that show is removed.
- [ ] Delete a whole downloaded anime from the library → confirm all of its cache entries and the entire `skipDetections[animeId]` entry are gone.
- [ ] Manually `rm` a `.mkv` file outside the app, restart → after the 30 s sweep, the cache entry for that file should be dropped (check console logs for `[skip-detector] cache sweep: kept N, dropped M stale entries`).
- [ ] Re-download an episode (different mtime) → after the next analysis, the old key should still be in cache; restart and verify the sweep clears it.
- [ ] Sanity: with a fresh app and clean cache, the sweep is a no-op (`dropped: 0`, no log).


src/main/index.ts
```diff
@@ -1628,6 +1628,131 @@ async function drainAutoSkipQueue(): Promise<void> {
...
+  return { kept, dropped }
+}
+
+function dropSkipDetectionsForAnime(animeId: number): boolean {
+  const detections = store.get('skipDetections') as Record<string, ShowSkipDetections>
+  const key = String(animeId)
+  if (!(key in detections)) return false
+  delete detections[key]
+  store.set('skipDetections', detections)
+  return true
```
mikkerlo, line 1741: nit: Since we already wired up `skip-detector:signature-updated` in Phase 3, we could broadcast it here so that if the user has `AnimeDetailView` or `PlayerView` open when they delete an anime, the UI live-updates and drops the OP/ED bounds without requiring a restart.

```suggestion
  store.set('skipDetections', detections)
  broadcastToAll('skip-detector:signature-updated', { animeId, perEpisode: {} })
  return true
```

src/main/index.ts
```diff
@@ -1628,6 +1628,131 @@ async function drainAutoSkipQueue(): Promise<void> {
...
+  const detections = store.get('skipDetections') as Record<string, ShowSkipDetections>
+  const key = String(animeId)
+  const show = detections[key]
+  if (!show || !show.perEpisode[episodeInt]) return false
+  delete show.perEpisode[episodeInt]
+  if (Object.keys(show.perEpisode).length === 0) {
+    delete detections[key]
+  }
+  store.set('skipDetections', detections)
+  return true
```
mikkerlo, line 1755: And similarly here for episode deletion:

```suggestion
  store.set('skipDetections', detections)
  broadcastToAll('skip-detector:signature-updated', {
    animeId,
    perEpisode: detections[key]?.perEpisode || {}
  })
  return true
```

mikkerlo (approved):
This is a great hygiene pass. Event-driven pruning paired with a startup backstop covers all the bases cleanly without polling. The synchronous `fs.statSync` inside the `setTimeout` backstop is totally fine since it's scanning a bounded local directory.

I just have one suggestion around the IPC broadcasts so that the UI can update live when an episode or anime is deleted.

LGTM after manual testing.

akvasha:
Thanks for the review! Both addressed in ed4026de:

1. **`dropSkipDetectionsForAnime` broadcast**: Fixed. Added `broadcastToAll('skip-detector:signature-updated', { animeId, perEpisode: {} })` after the store write so an open `AnimeDetailView` / `PlayerView` re-fetches and drops the bands without a restart.
2. **`dropSkipDetectionsForEpisode` broadcast**: Fixed. Same channel; payload sends the *remaining* `perEpisode` after the deletion (or `{}` if the show became empty and the parent key was removed too) so the receiver sees the updated state directly.

Typecheck clean.

Issue #63, opened 2026-05-02
Support watching episodes while downloading


mikkerlo:
**Priority:** Medium | **Effort:** Large

## Motivation
Right now, downloading and watching are disconnected. Users want to start watching an episode that is actively downloading from the local `.part` file. This saves bandwidth compared to streaming it from the CDN again while it's also downloading. It provides a smoother experience since local playback won't buffer if the download is sufficiently ahead.

## Approach & Architecture
We will update the download manager to prioritize downloading `.ass` subtitle files before the `.mp4` video. The local custom protocol handler (`anime-video://`) will be modified to serve growing `.part` files, querying the `DownloadManager` for the expected `totalBytes` to spoof the `Content-Length`. The player will be updated to accept `.part` files, clamp seeking based on download progress, and show a buffering spinner when the playhead catches up to the download. Merging of the downloaded file will be deferred until the user closes the player to avoid locking conflicts.

## UI / UX Considerations (If applicable)
- The Player UI will need to visually represent download progress on the seek bar (e.g., using a different colored track or the native buffered track).
- A buffering spinner must appear when playback catches up to the downloaded boundary.
- Seeking should be clamped to only allow jumping to portions that have already been downloaded.
- The user flow shouldn't change significantly; the "Play" button on a downloading episode will just open the local player instead of the stream.

## Implementation Plan

1. **Main Process (`src/main/`):**
   - [ ] Update `src/main/download-manager.ts` enqueue/processing logic to prioritize subtitle (`.ass`) files over video (`.mp4`) files.
   - [ ] Modify `anime-video://` protocol handler in `src/main/index.ts` to support `.part` files. It should query `DownloadManager` for `totalBytes` to set `Content-Length` and `Content-Range`.
   - [ ] Ensure the protocol handler handles read streams reaching current EOF properly (e.g., return short 206 responses).
   - [ ] Track player lock state in `src/main/index.ts` and `DownloadManager`. If a download finishes while locked, set state to `deferred`.
   - [ ] Trigger merge process for deferred downloads when the player is closed (if auto-merge is enabled).
2. **Preload & IPC (`src/preload/`):**
   - [ ] Add `player:opened` and `player:closed` IPC endpoints in `preload/index.ts` and `preload/types.d.ts` to track file locks.
3. **Renderer (`src/renderer/`):**
   - [ ] Update `PlayerView.vue` to accept `.part` files.
   - [ ] Subscribe to `download:progress` IPC events for the active episode.
[...]

PR #67, opened 2026-05-04
Probe MP4 faststart on download and open

mikkerlo:
## Summary
- Adds a small MP4 box-header walker (`src/main/mp4-faststart.ts`) that reads the first ~64 KB of a file and reports whether `moov` precedes `mdat`.
- Runs the probe in two places:
  - When a video download finalizes (`.part → .mp4` rename), via a new `onVideoDownloaded` callback on `DownloadManager`.
  - When the player resolves a local MP4 via `player:find-local-file` (so users with already-downloaded files contribute samples too).
- Persists stats in electron-store (`mp4StreamingStats`) — total checked, faststart count, and up to 10 most-recent non-faststart samples (anime + episode + path).
- Shows the stats on the existing Debug tab in Settings, with the most recent non-faststart example and a Reset button.
- An in-memory per-session `Set<string>` keeps the player path from double-counting when the same file is opened multiple times.
- Patch version bump to 3.19.1.
- MKV files are explicitly skipped — both call sites guard on `.mp4` suffix, so the MKV/MSE playback path is untouched.

## Why
Issue #63 (watch episodes while downloading) only works if the CDN serves streaming-optimized MP4s (`moov`-at-start). A spike against one episode confirmed the primary CDN does, but we want background telemetry before that feature ships, so a different show or quality variant served by a non-faststart encoder doesn't surprise us in production.

If a non-faststart file shows up in the wild, the debug panel surfaces a concrete anime + episode + path so it can be reproduced and investigated.

## Test plan
- [ ] Download a fresh episode, confirm `Faststart: 1 / 1` on the Debug tab.
- [ ] Open an already-downloaded MP4 in the built-in player, confirm the count increments.
- [ ] Open the same MP4 again — count should not change (per-session de-dup).
- [ ] Verify MKV opens do not affect the count.
- [ ] Verify stats survive an app restart.

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -31,6 +31,8 @@ import type {
   SyncplayRemoteEpisode,
   SyncplayStatus
 } from './syncplay'
+import { probeMp4Faststart } from './mp4-faststart'
+import type { Mp4StreamingStats } from './mp4-faststart'
 
 // Enable WebGPU (Anime4K shaders) and platform HEVC decoding (HEVC MKV via MSE).
 // PlatformHEVCDecoderSupport gates Chromium's HEVC path in <video> and MSE;
@@ -122,7 +124,12 @@ const store = new Store({
       lastRoom: string
       username: string
       autoReconnect: boolean
-    }
+    },
+    mp4StreamingStats: {
+      totalChecked: 0,
+      faststartCount: 0,
+      nonFaststartSamples: [] as { animeId: number; animeName: string; episodeInt: string; episodeLabel: string; filePath: string; firstNonFtypBox: string; checkedAt: number }[]
+    } as Mp4StreamingStats
   }
 })
 
@@ -162,6 +169,42 @@ function isNetworkError(err: unknown): boolean {
   const code = errorCode(err)
   if (code && NETWORK_ERROR_CODES.has(code)) return true
   return false
+}
+
+// In-memory set of MP4 paths probed this session, so re-opening the same file
+// in the player doesn't double-count. Stats persist across sessions; this set
+// resets on app restart, which is fine — it's a sampling check, not an exact
+// per-file ledger.
+const mp4FaststartChecked = new Set<string>()
+
+async function recordMp4FaststartCheck(
+  filePath: string,
+  context: { animeId: number; animeName: string; episodeInt: string; episodeLabel: string }
+): Promise<void> {
+  if (mp4FaststartChecked.has(filePath)) return
+  mp4FaststartChecked.add(filePath)
+  const probe = await probeMp4Faststart(filePath)
+  if (!probe) return
+  const stats = store.get('mp4StreamingStats') as Mp4StreamingStats
+  stats.totalChecked += 1
+  if (probe.faststart) {
+    stats.faststartCount += 1
+  } else {
+    stats.nonFaststartSamples.push({
+      animeId: context.animeId,
+      animeName: context.animeName,
+      episodeInt: context.episodeInt,
+      episodeLabel: context.episodeLabel,
+      filePath,
+      firstNonFtypBox: probe.firstNonFtypBox,
+      checkedAt: Date.now()
+    })
+    if (stats.nonFaststartSamples.length > 10) {
+      stats.nonFaststartSamples = stats.nonFaststartSamples.slice(-10)
+    }
+    console.error(`[mp4-faststart] non-faststart MP4 detected: ${context.animeName} — ${context.episodeLabel} (first non-ftyp box: ${probe.firstNonFtypBox}) at ${filePath}`)
+  }
+  store.set('mp4StreamingStats', stats)
 }
 
 interface QueuedShikimoriUpdate {
@@ -1891,6 +1934,14 @@ function registerIpcHandlers(): void {
     return qualityMismatches.size
   })
 
+  ipcMain.handle('debug:get-mp4-stats', () => {
+    return store.get('mp4StreamingStats') as Mp4StreamingStats
+  })
+
+  ipcMain.handle('debug:reset-mp4-stats', () => {
+    store.set('mp4StreamingStats', { totalChecked: 0, faststartCount: 0, nonFaststartSamples: [] })
+  })
+
   ipcMain.handle('dump-quality-mismatches', () => {
     const outPath = path.join(getDownloadDir(), 'quality-mismatches.json')
     const data = [...qualityMismatches.values()]
@@ -3206,6 +3257,17 @@ function registerIpcHandlers(): void {
       if (coldDir) dirsToCheck.push(coldDir)
     }
 
+    const onResolved = (fp: string): void => {
+      if (fp.toLowerCase().endsWith('.mp4')) {
+        void recordMp4FaststartCheck(fp, {
+          animeId: 0,
+          animeName,
+          episodeInt,
+          episodeLabel: `Ep ${episodeInt}`
+        })
+      }
+    }
+
     for (const dir of dirsToCheck) {
       const animeDir = path.join(dir, animeDirName)
       // Try tagged filename first
@@ -3216,6 +3278,7 @@ function registerIpcHandlers(): void {
             const assPath = fp.replace(/\.(mp4|mkv)$/i, '.ass')
             try { return fs.existsSync(assPath) ? fs.readFileSync(assPath, 'utf-8') : null } catch { return null }
           })()
+          onResolved(fp)
           return { filePath: fp, subtitleContent }
         }
       }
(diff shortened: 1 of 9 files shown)


src/main/index.ts
```diff
@@ -141,6 +148,42 @@ function isNetworkError(err: unknown): boolean {
...
+      episodeInt: context.episodeInt,
+      episodeLabel: context.episodeLabel,
+      filePath,
+      firstNonFtypBox: probe.firstNonFtypBox,
+      checkedAt: Date.now()
+    })
+    if (stats.nonFaststartSamples.length > 10) {
+      stats.nonFaststartSamples = stats.nonFaststartSamples.slice(-10)
+    }
+    console.error(`[mp4-faststart] non-faststart MP4 detected: ${context.animeName} — ${context.episodeLabel} (first non-ftyp box: ${probe.firstNonFtypBox}) at ${filePath}`)
```
akvasha: nit: this is non-fatal telemetry, not an error condition — `console.warn` matches the rest of the codebase and avoids surfacing it in error reporters.

```suggestion
    console.warn(`[mp4-faststart] non-faststart MP4 detected: ${context.animeName} — ${context.episodeLabel} (first non-ftyp box: ${probe.firstNonFtypBox}) at ${filePath}`)
```

src/main/index.ts
```diff
@@ -1868,6 +1911,14 @@ function registerIpcHandlers(): void {
     return qualityMismatches.size
   })
 
+  ipcMain.handle('debug:get-mp4-stats', () => {
+    return store.get('mp4StreamingStats') as Mp4StreamingStats
+  })
+
+  ipcMain.handle('debug:reset-mp4-stats', () => {
+    store.set('mp4StreamingStats', { totalChecked: 0, faststartCount: 0, nonFaststartSamples: [] })
```
akvasha, line 1950: nit: Reset clears persisted stats but leaves the in-memory `mp4FaststartChecked` set populated, so re-opening the same file in the player after Reset is a no-op until app restart. Consider also clearing the set here:

```suggestion
    store.set('mp4StreamingStats', { totalChecked: 0, faststartCount: 0, nonFaststartSamples: [] })
    mp4FaststartChecked.clear()
```

src/main/index.ts
```diff
@@ -3183,6 +3234,17 @@ function registerIpcHandlers(): void {
       if (coldDir) dirsToCheck.push(coldDir)
     }
 
+    const onResolved = (fp: string): void => {
+      if (fp.toLowerCase().endsWith('.mp4')) {
+        void recordMp4FaststartCheck(fp, {
+          animeId: 0,
```
akvasha, line 3276: nit: hardcoded `animeId: 0` means non-faststart samples surfaced from the player path can't be cross-referenced by id later. Not blocking — `recentAnimeMeta` is keyed by id rather than name, so a clean lookup isn't trivial — but worth a short comment noting why it's 0 for future readers.

akvasha (approved):
Telemetry-only and well-scoped — sensible to land before #63 phase 2.

One project-convention follow-up: DESIGN.md should be updated for the two new `debug:*-mp4-stats` IPC handlers and the new `mp4StreamingStats` electron-store key. Otherwise just a couple of nits inline.

LGTM after manual testing.

mikkerlo:
Thanks for the review! Pushed fixes for all three:

1. **`console.error` → `console.warn`** for non-faststart detection (`src/main/index.ts:182`). Agreed it's telemetry, not an error.
2. **Reset clears `mp4FaststartChecked` set** (`src/main/index.ts:1919-1922`). Without this, re-opening the same files after Reset was a no-op until restart, defeating the point of clicking Reset to redo the check.
3. **Comment added on hardcoded `animeId: 0`** (`src/main/index.ts:3237-3242`). Explains that the IPC handler only receives `animeName`; resolving back to an id would require scanning `recentAnimeMeta`, and since the sample is for human inspection (title + episode + path), the missing id is acceptable.

src/main/index.ts
```diff
@@ -3183,6 +3235,21 @@ function registerIpcHandlers(): void {
...
+      if (fp.toLowerCase().endsWith('.mp4')) {
+        // animeId is 0 here because this handler only receives animeName; resolving
+        // back to an id would require scanning recentAnimeMeta. The stats sample is
+        // for human inspection (anime title + episode + filepath), so the missing
+        // id is acceptable.
+        void recordMp4FaststartCheck(fp, {
+          animeId: 0,
+          animeName,
+          episodeInt,
+          episodeLabel: `Ep ${episodeInt}`
```
akvasha: nit: the download path stores the real `item.episodeLabel` (e.g. `"Episode 5 (AniLibria)"`), but this path synthesizes `"Ep 5"`. Samples surfaced in the Settings panel will mix the two formats depending on which path produced them. Could fix by extending the IPC signature to take `episodeLabel`, or by reading it back from `downloadedEpisodes` (currently only stores `author/quality/translationId`, would need a small schema bump). Not blocking — the panel is for human inspection.

src/main/index.ts
```diff
@@ -141,6 +148,42 @@ function isNetworkError(err: unknown): boolean {
...
+      filePath,
+      firstNonFtypBox: probe.firstNonFtypBox,
+      checkedAt: Date.now()
+    })
+    if (stats.nonFaststartSamples.length > 10) {
+      stats.nonFaststartSamples = stats.nonFaststartSamples.slice(-10)
+    }
+    console.warn(`[mp4-faststart] non-faststart MP4 detected: ${context.animeName} — ${context.episodeLabel} (first non-ftyp box: ${probe.firstNonFtypBox}) at ${filePath}`)
+  }
+  store.set('mp4StreamingStats', stats)
```
akvasha: nit: this is non-atomic — two video downloads finalizing at nearly the same time will both `store.get` deep-clones, both mutate, and the second `store.set` clobbers the first's increment. Off-by-one on the counter only, harmless for telemetry, but worth noting in case this ever gets reused for something stricter.

src/renderer/src/components/SettingsView.vue
```diff
@@ -1458,6 +1477,27 @@ watch(autoCleanupDays, (val) => { if (loaded.value) autoSave('autoCleanupWatched
             <div class="scan-error-item">{{ dumpResult.path }}</div>
           </div>
         </div>
+
+        <div class="setting-group">
+          <label class="setting-label">MP4 streaming-optimization check</label>
+          <p class="setting-hint">After every video download or when an MP4 is opened in the player, the first ~64 KB are scanned for the MP4 box order. Streaming-optimized files (`moov` before `mdat`) are required to play while still downloading. If a non-faststart file is found, an example is shown below for inspection.</p>
```
akvasha: nit: the backticks around `moov` / `mdat` render literally in the panel since this is plain HTML, not Markdown. Either drop them or wrap with `<code>`:

```suggestion
          <p class="setting-hint">After every video download or when an MP4 is opened in the player, the first ~64 KB are scanned for the MP4 box order. Streaming-optimized files (<code>moov</code> before <code>mdat</code>) are required to play while still downloading. If a non-faststart file is found, an example is shown below for inspection.</p>
```

akvasha (approved):
Previous round of nits all addressed cleanly. On a fuller pass I noticed a couple more small things — see inline. None blocking.

Still outstanding from the prior review: DESIGN.md update for the new IPC handlers (`debug:get-mp4-stats`, `debug:reset-mp4-stats`) and the `mp4StreamingStats` store key. Not blocking but worth a follow-up.

LGTM after manual testing.

mikkerlo:
Round-2 fixes pushed:

1. **SettingsView backticks** (`src/renderer/src/components/SettingsView.vue:1663`). Wrapped `moov` / `mdat` in `<code>` tags so they render as code in the Settings panel (the hint is plain HTML, not Markdown).
2. **`episodeLabel` synthesis** (`src/main/index.ts` + IPC). Extended `player:find-local-file` to accept the friendly `episodeLabel` from the renderer; PlayerView now passes `episodeFull` (or the active label as a fallback). Non-faststart samples surfaced from the player path will now match the labeling from the download path instead of mixing `"Ep 5"` with `"Episode 5 (AniLibria)"`.
3. **Non-atomic stats write** (`src/main/index.ts:178-218`). Added a promise-chain serializer (`mp4StatsWriteChain`) around the `store.get` → mutate → `store.set` so two near-simultaneous download finishes can't clobber each other's increments. Cheap to add and removes the off-by-one even though it was harmless for telemetry.
4. **DESIGN.md** (carryover from both review summaries). Documented the `debug:get-mp4-stats` / `debug:reset-mp4-stats` IPC channels, the `mp4StreamingStats` electron-store key, and the new `episodeLabel` parameter on `player:find-local-file`.

PR #68, opened 2026-05-05
feat: remember selected translation per episode

### LINKED ISSUE #50: Remember Selected Translation per Episode

## Motivation
If a user changes the translation for an episode (either in the episode list or inside the player), the app should remember the specific translation when they return, instead of reverting to the global default. It must save the choice immediately upon switching. An explicitly remembered translation must override a downloaded file in priority to ensure the user gets exactly what they last selected.

## Approach & Architecture
Add an optional `translationId` to the `WatchProgressEntry` stored via `electron-store`. The episode list view (pre-player dropdown) and the player will save the chosen translation to the progress entry upon change. The episode list view will look up this ID as a preferred choice. To resolve issues with `props.translationId` being static, we will use a reactive `activeTranslationId` to track the current translation within the player.

No new UI elements. The user will simply notice their translation choice is persisted for an episode if they leave and return. Pre-player dropdown changes MUST also save persistently. An explicitly remembered stream translation MUST override a downloaded file in priority.

## Implementation Plan

   - [ ] Update `watch-progress:save` handler in `src/main/index.ts` to accept and persist `translationId` in `WatchProgressEntry`.
   - [ ] Add `translationId?: number` to `WatchProgressEntry` interface in `src/preload/types.d.ts`.
   - [ ] Update `watchProgressSave` signature in `src/preload/index.ts` and `src/preload/types.d.ts` to accept `translationId?: number`.
   - [ ] In `PlayerView.vue`, replace usages of `props.translationId` with a reactive `activeTranslationId` (initialized from props or progress) to correctly track state changes.
   - [ ] In `PlayerView.vue`, pass `activeTranslationId.value` to `window.api.watchProgressSave` inside `saveProgress()` and `markWatched()`.
   - [ ] In `PlayerView.vue`, immediately call `window.api.watchProgressSave` with the new translation ID upon translation change (e.g., in `onTranslationChange` or a watcher).
   - [ ] In `AnimeDetailView.vue`, update `selectedTr` initialization in `episodeRows`. Insert a new highest priority level (Priority 1) to check `watchProgress.value[ep.episodeInt]?.translationId`. If present and valid for the episode, use it, explicitly overriding downloaded files.
   - [ ] In `AnimeDetailView.vue`'s translation dropdown change handler, call `window.api.watchProgressSave` to persist pre-player dropdown changes immediately.

## Files to Touch

## Testing Strategy
- Change the translation in the episode list dropdown, navigate away, and return. It should persist.
- Open an episode and change the translation, then exit the player.
- Re-open the same episode; it should resume with the previously chosen translation.
- Verify that if a local downloaded file exists, but the user explicitly selected a different stream translation previously, the remembered stream translation takes precedence over the local download.
- Test normal watch progress saves still work without regressions.

## Risks & Edge Cases
- **Stale Translation ID:** The remembered `translationId` may no longer be available in the episode's translation list; must fallback gracefully to downloads or global defaults.
- **Save Frequency:** Calling save immediately upon translation change shouldn't conflict with normal progress polling saves.
- **Downloaded precedence:** Overriding downloaded translations means users might accidentally stream when they have a local file if they had previously selected a stream. The UI should remain clear.

## Out of Scope
- Remembering video quality, subtitle track, or audio track preferences.
- Changing the global default translation preference.

mikkerlo:
## Summary

Persist the active translation id into `watchProgress` so reopening an episode preselects the last-chosen translation. The remembered choice overrides downloaded files — an explicit user pick (even a stream) takes precedence over what's on disk, per the updated requirements in #50.

- `watchProgress` entries gain `translationId`; the save handler preserves the prior value when not passed and overwrites when a value is passed.
- `PlayerView.saveProgress` now passes `activeTranslationId`. A new `persistSelectedTranslation` helper fires immediately on `selectTranslation` (both local-file and stream paths), bypassing the trivial-time guard so the choice survives even a quick switch-then-close.
- `AnimeDetailView.onEpisodeTranslationChange` also persists via `watchProgressSave` (reusing prior `position`/`duration` so partial-watch progress isn't clobbered when only the translation changes).
- `AnimeDetailView.episodeRows` resolves `selectedTr` in the order: queue lock → per-episode override (in-session, sync) → **remembered translationId** → any downloaded translation → global type+author default. Queue-lock stays at the top because the row UI replaces the dropdown with a "Queued" label while a download is in flight.

Fixes #50

## Test plan

- [ ] Open an episode in the built-in player; switch to a non-default translation; close the player. Reopen the same episode from `AnimeDetailView` — the same translation is preselected on the row and in the player.
- [ ] Download a file in a *different* translation than the remembered one; verify the **remembered translation still wins** (it now overrides downloaded files).
- [ ] Change the translation via the episode-list dropdown (without opening the player); navigate away and back — the choice persists.
- [ ] An in-flight download still locks the row to the queued translation (the dropdown is replaced with a \"Queued\" label).
- [ ] Marking a previous episode as watched (auto-advance scenario) does not clobber that episode's remembered \`translationId\`.
- [ ] \`npm run typecheck\` clean.

### DIFF
--- a/src/renderer/src/components/PlayerView.vue
+++ b/src/renderer/src/components/PlayerView.vue
@@ -467,10 +467,24 @@ async function saveProgress(force = false): Promise<void> {
   // When watched, clear the position so we don't try to resume near the end later
   const positionToSave = watchedReported ? 0 : video.currentTime
   try {
-    await window.api.watchProgressSave(props.animeId, epInt, positionToSave, duration.value, watchedReported)
+    await window.api.watchProgressSave(props.animeId, epInt, positionToSave, duration.value, watchedReported, activeTranslationId.value ?? undefined)
     window.dispatchEvent(new CustomEvent('watch-progress-updated'))
   } catch (err) {
     console.warn('[player] failed to save watch progress:', err)
+  }
+}
+
+async function persistSelectedTranslation(translationId: number): Promise<void> {
+  const epInt = currentEpisodeInt.value
+  if (!props.animeId || !epInt) return
+  const video = videoRef.value
+  const pos = watchedReported ? 0 : (video?.currentTime ?? 0)
+  const dur = duration.value || (video?.duration ?? 0)
+  try {
+    await window.api.watchProgressSave(props.animeId, epInt, pos, dur, watchedReported, translationId)
+    window.dispatchEvent(new CustomEvent('watch-progress-updated'))
+  } catch (err) {
+    console.warn('[player] failed to persist translation choice:', err)
   }
 }
 
@@ -1662,6 +1676,7 @@ async function selectTranslation(tr: { id: number; label: string; type: string; 
       const localResult = await window.api.playerFindLocalFile(props.animeName, activeEpisodeLabel.value, tr.id, friendlyLabel)
       if (localResult) {
         activeTranslationId.value = tr.id
+        persistSelectedTranslation(tr.id)
 
         // Clean up previous remux / stream session if any
         if (remuxedPath.value || streamSessionId.value) {
@@ -1710,6 +1725,7 @@ async function selectTranslation(tr: { id: number; label: string; type: string; 
     }
 
     activeTranslationId.value = tr.id
+    persistSelectedTranslation(tr.id)
 
     // Clean up previous remux / MSE stream if switching from local to stream
     if (remuxedPath.value || streamSessionId.value) {
--- a/src/renderer/src/components/PlayerView.vue
+++ b/src/renderer/src/components/PlayerView.vue
@@ -478,8 +478,19 @@ async function persistSelectedTranslation(translationId: number): Promise<void> 
   const epInt = currentEpisodeInt.value
   if (!props.animeId || !epInt) return
   const video = videoRef.value
-  const pos = watchedReported ? 0 : (video?.currentTime ?? 0)
-  const dur = duration.value || (video?.duration ?? 0)
+  const vidDur = video?.duration && !Number.isNaN(video.duration) ? video.duration : 0
+  let pos = watchedReported ? 0 : (video?.currentTime ?? 0)
+  let dur = duration.value || vidDur
+  if (!dur) {
+    // Pre-loadedmetadata switch: avoid clobbering existing resume position with 0/0
+    try {
+      const prev = await window.api.watchProgressGet(props.animeId, epInt)
+      if (prev) {
+        pos = prev.watched ? 0 : prev.position
+        dur = prev.duration
+      }
+    } catch { /* ignore */ }
+  }
   try {
     await window.api.watchProgressSave(props.animeId, epInt, pos, dur, watchedReported, translationId)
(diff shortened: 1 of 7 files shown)


mikkerlo:
### Nits & Clarifications

- **Summary Discrepancy**: The PR summary mentioned that downloaded files win over remembered streams, but the code correctly implements the opposite (remembered wins), which aligns with the requirements in #50. This appears to be a minor typo in the PR description.
- **Type Safety**: Interface updates in `src/preload/types.d.ts` correctly ensure full type safety for the new field across the bridge.

mikkerlo:
Thanks for the review!

1. **Summary Discrepancy** — Fixed. The PR description was stale from the first pass, before the alignment commit (`24e5bdcd`) reordered the priority list. Updated the body to match the current behavior: remembered translation overrides downloaded files. Also corrected the test plan and added a check for the new dropdown-persists-too path. Good catch.
2. **Type Safety** — Acknowledged, thanks.

src/renderer/src/components/PlayerView.vue
```diff
@@ -467,13 +467,27 @@ async function saveProgress(force = false): Promise<void> {
...
     console.warn('[player] failed to save watch progress:', err)
   }
 }
 
+async function persistSelectedTranslation(translationId: number): Promise<void> {
+  const epInt = currentEpisodeInt.value
+  if (!props.animeId || !epInt) return
+  const video = videoRef.value
+  const pos = watchedReported ? 0 : (video?.currentTime ?? 0)
+  const dur = duration.value || (video?.duration ?? 0)
```
akvasha: nit: if the user switches translation before `loadedmetadata` fires (slow CDN open is the realistic case), both `duration.value` and `video?.duration` are still 0 here — so we'll persist `position: 0, duration: 0` and silently clobber any saved resume position for the episode. `saveProgress` sidesteps this by early-returning on `!duration.value`, but here we still want to write the new `translationId`. One option is to fetch `prev` via `watchProgressGet` when duration is unknown and reuse its position/duration:

```suggestion
  const dur = duration.value || (video?.duration && !Number.isNaN(video.duration) ? video.duration : 0)
  let pos = watchedReported ? 0 : (video?.currentTime ?? 0)
  let durToSave = dur
  if (!dur) {
    try {
      const prev = await window.api.watchProgressGet(props.animeId, epInt)
      if (prev) {
        pos = prev.watched ? 0 : prev.position
        durToSave = prev.duration
      }
    } catch { /* ignore */ }
  }
```

akvasha (approved):
LGTM after manual testing.

mikkerlo:
1. **Pre-metadata translation switch** — Fixed in `5f6ae97`. `persistSelectedTranslation` now falls back to `watchProgressGet` and reuses the prior `position`/`duration` when both `duration.value` and `video.duration` are still 0, so a switch before `loadedmetadata` fires no longer clobbers the saved resume point. Also guards against `NaN` from `video.duration` while we're at it.

PR #71, opened 2026-05-06
fix: frame-accurate seeking in built-in MKV player

### LINKED ISSUE #65: Fix imprecise seeking in video player

## Motivation
When seeking to an unbuffered region (or even within buffered regions) in the video player, the playback often lands a few seconds before the intended spot. 

This happens because of how MSE (Media Source Extensions) and FFmpeg are synchronized. Currently, the player calculates a seek time `seekAt = target - 1`, and tells FFmpeg to start at `seekAt`. FFmpeg seeks to the nearest keyframe at or before `seekAt`, resetting timestamps to 0. The player then manually maps this to `seekAt` in the player's timeline, shifting the entire timeline backwards and making seeking imprecise.

## Approach & Architecture
Instead of having FFmpeg force timestamps to zero and manually calculating offsets in the renderer, we will instruct FFmpeg to preserve original timestamps and normalize them to start at 0 relative to the file start. The renderer will then perform simple timeline alignment by setting a constant zero offset, treating the incoming stream's timestamps as the absolute truth.

No direct visual changes to the UI, but user experience will be significantly improved. Seeking will become frame-accurate, and the playhead will no longer unexpectedly jump backwards.

## Implementation Plan

   - [ ] In `src/main/index.ts` (`spawnFfmpegForSession`): Add `-copyts` and `-start_at_zero` to the FFmpeg arguments to preserve absolute timestamps.
   - [ ] In `src/main/index.ts` (`spawnFfmpegForSession`): Remove `-avoid_negative_ts make_zero`.

   - *No changes required.*

   - [ ] In `src/renderer/src/components/PlayerView.vue`: Set `sourceBuffer.timestampOffset = 0` unconditionally (e.g., in `startMseSession` and `handleUnbufferedSeek`).
   - [ ] In `src/renderer/src/components/PlayerView.vue` (`handleUnbufferedSeek`): Ensure the `<video>` element is seeked to the exact `target` requested, even if FFmpeg is spawned at `target - 1`.
   - [ ] In `src/renderer/src/components/PlayerView.vue`: Update `mseInitialSeek` logic to reflect the intended `target` time so `resumeFromSavedPosition` lands precisely on the mark.

## Files to Touch

## Testing Strategy
- **Unbuffered Seeking:** Seek to various positions (especially the middle of a long GOP) and verify the `currentTime` matches the intended position exactly.
- **Backward Seeking:** Verify that seeking backwards to unbuffered regions works correctly and lands accurately.
- **Buffered Seeking:** Verify that subsequent seeks within the newly buffered regions remain accurate without shifting.
- **Resume:** Verify resuming playback from a saved position starts at the exact correct second.
- **Transcode Path:** Verify that the transcode path (e.g., HEVC fallback) still functions correctly and benefits from better sync.

## Risks & Edge Cases
- **Transcode Compatibility:** `-copyts` combined with `-start_at_zero` must not break the A/V synchronization when active transcoding (e.g., HEVC -> H264) is happening.
- **Timestamp Gaps:** Some malformed video files might have timestamp gaps; preserving timestamps might propagate these gaps to the player.
- **Negative Timestamps:** Need to ensure `-start_at_zero` effectively handles files that might natively start with negative timestamps, preventing MSE decoding errors.

## Out of Scope
- Complete refactoring of the MSE player session management.
- Modifying the static 1-second pre-buffer (`target - 1`) logic; this issue only focuses on how timestamps within that buffer are mapped to the timeline.

akvasha:
## Summary

- **Frame-accurate MSE seeks.** ffmpeg's fmp4 muxer normalizes its `tfdt.baseMediaDecodeTime` to start at 0 per output (even with `-copyts`), so the renderer can't infer file-time from output PTS. Main now probes the actual keyframe at-or-before each seek with `ffprobe -skip_frame nokey -read_intervals` over a 15 s window, hands the exact keyframe time back to the renderer, and the renderer sets `sourceBuffer.timestampOffset = keyframeTime`. The buffered range then maps onto the absolute file timeline and `<video>.currentTime` lands on the user's exact target instead of the keyframe.
- **Seek-bar tooltip vs click alignment.** The default `<input type="range">` thumb (~16 px) clamps the input's value mapping to `[thumb/2, width − thumb/2]`, while the hover tooltip used linear `x / width × duration`. Adding a zero-width `::-webkit-slider-thumb` / `::-moz-range-thumb` makes both maps linear so clicking exactly where the tooltip says lands you there.

Fixes #65

## Test plan

- [x] Resume from a saved position lands within container precision of the saved second (no ±GOP offset).
- [x] Forward seek into an unbuffered region: tooltip time matches the actual playhead and visible content.
- [x] Backward seek into an unbuffered region: same.
- [ ] HEVC transcode path still seeks correctly.
- [x] Unbuffered seek no longer triggers a spurious `ended` / auto-advance.


mikkerlo (approved):
The approach of probing for the exact keyframe and offsetting the `SourceBuffer` timeline is a clever and robust solution to the fmp4 timestamp normalization issue. The CSS fixes for the seek slider are also a great touch for improving precision. LGTM after manual testing

PR #72, opened 2026-05-06
feat: cache-first AnimeDetailView load with 24h TTL

### LINKED ISSUE #66: Optimize AnimeDetailView Load Time

## Motivation
Currently, the `AnimeDetailView` waits for all episode translations and file statuses to load before dropping the global loading screen. This creates a slow perceived load time. By caching full anime details locally for starred or downloaded anime (with a 1-day TTL), we can render the page instantly and fetch fresh updates in the background, significantly improving perceived performance.

## Approach & Architecture
Instead of relying on piecemeal optimistic UI updates, we will introduce a caching layer for full anime details.
- If an anime is starred or has downloaded episodes, its full details will be cached locally (TTL 1 day).
- When opening an anime, if a valid cache exists, it will be used to render the page instantly.
- A background fetch will run to update the data; once the fetch resolves, the UI and cache will be updated with the fresh data.

## UI / UX Considerations
- The "spinner over blank screen" will be replaced by instant rendering for cached (starred/downloaded) anime.
- For uncached anime, a global loading indicator will still show until the initial fetch completes.
- The UI must handle background data refreshes gracefully without causing disruptive flickering.

## Implementation Plan

   - [ ] Implement a caching mechanism for anime details. Store the full response payload along with a timestamp.
   - [ ] Add an IPC handler `api.getAnimeCache(id)` to retrieve cached data if it exists and is within the 1-day TTL.
   - [ ] Add an IPC handler `api.setAnimeCache(id, data)` to store data when fetched, but only if the anime is starred or has downloaded episodes.
   - [ ] Add `getAnimeCache` and `setAnimeCache` to `preload/index.ts` and `preload/types.d.ts`.
   - [ ] In `AnimeDetailView.vue`, modify the initialization logic to first attempt loading from `api.getAnimeCache(id)`.
   - [ ] If a cache hits, set `anime.value = cachedData`, drop the loading spinner immediately, and render the page.
   - [ ] Fire a background fetch using `api.getAnime(id)`. When it returns, update `anime.value` and call `api.setAnimeCache(id, newData)`.
   - [ ] Implement a generation counter or `disposed = true` flag in `AnimeDetailView.vue`'s `onUnmounted` to prevent state pollution if the user rapidly switches anime.

## Files to Touch
- **Documentation:** `DESIGN.md`

## Testing Strategy
- Open a starred anime, verify it loads instantly from cache.
- Wait for the background fetch to complete and verify the UI updates seamlessly.
- Switch rapidly between different anime to ensure no state pollution.
- Validate the 1-day TTL logic by manually modifying the cache timestamp and verifying it acts as a miss.
- Test with an uncached (not starred, not downloaded) anime to ensure normal loading behavior.

## Risks & Edge Cases
- Background refresh mutating `anime.value` might cause UI flicker or interrupt user interaction (e.g., selecting a translation).
- Race conditions during rapid anime switching while background fetches are resolving.

## Out of Scope
- Adding full 429 backoff/retry support for the API.

akvasha:
## Summary

Eliminates the multi-second blank state when opening a starred or downloaded anime. The detail view now renders instantly from a local cache (24h TTL) and refreshes in the background.

Fixes #66.

## What changed

**Main process (`src/main/index.ts`)**
- `isCachableAnime(id)` = downloaded OR starred. Detail / episode / quality-probe caches are now gated on this instead of "downloaded only".
- `cleanupStaleCache` keeps entries that are downloaded *or* in the library.
- Unstarring a non-downloaded anime evicts its cache entry.
- New IPC: `get-anime-cache` (returns `{ data, cachedAt }` if fresh, else `null`) and `set-anime-cache` (writes only if cachable).

**Preload (`src/preload/index.ts`, `types.d.ts`)**
- `window.api.getAnimeCache(id)` and `setAnimeCache(id, data)`.

**Renderer (`AnimeDetailView.vue`)**
- On mount: read cache → if fresh, render immediately and load episodes / file status. Then fire the live `get-anime` request and reconcile.
- Module-scoped `loadGeneration` counter + `disposed` flag guard against rapid anime switching: late responses from the previous anime are dropped instead of clobbering the new view.
- Extracted the translation-default logic into `applyDownloadedTranslationDefault()` so the cache-render and live-render paths share it.

**Docs**
- `DESIGN.md` updated with the new IPC handlers and the cache-first flow.

## Test plan

- [x] `npm run typecheck` clean (node + web)
- [x] Verified locally: opening a starred anime renders instantly; switching rapidly between anime does not leak stale data into the new view
- [ ] Verify on a fresh install that a never-opened anime still goes through the live path (no cache yet)
- [ ] Verify unstarring a non-downloaded anime evicts its cache (next open should hit the API)

## Out of scope

The Shikimori panel (`get-rate`, `get-friends-rates`, `get-related`) is still uncached and remains the slow part of the page. See follow-up plan in the comment below.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -98,7 +98,8 @@ translation type and author selections across re-mounts.
 ### Episode Loading & Translation Selection
 
 ```
-AnimeDetailView  -->  getAnime(id) --> full anime with episode list
+AnimeDetailView  -->  getAnimeCache(id) --> instant render if cached and fresh
+                 -->  getAnime(id) --> full anime with episode list (background-refresh)
                  -->  getEpisode(id) x N (batches of 5) --> translations per episode
                  -->  all active translations shown (no quality minimum)
                  -->  best quality per author+type deduplication
@@ -108,6 +109,15 @@ AnimeDetailView  -->  getAnime(id) --> full anime with episode list
 Episode list is paginated (30 per page) when anime has >30 TV episodes.
 Only translations for the visible page are fetched (in batches of 5).
 Cached translations are reused when revisiting a page.
+
+For starred or downloaded anime, the full AnimeDetail payload is cached in
+`animeCache[animeId].animeDetail` with a 24h TTL. AnimeDetailView calls
+`getAnimeCache(id)` on mount: on a hit, the page renders instantly from the
+cache while a background `getAnime(id)` refreshes the data; on miss, the
+spinner stays up until the API responds. A per-mount generation counter +
+`disposed` flag prevent stale background fetches from polluting state when
+the user rapidly switches anime. Cache writes are gated on starred OR
+downloaded; unstarring evicts the cache entry if the anime isn't downloaded.
 
 Per-episode translation selector with priority chain:
   1. In download queue → locked to queued translation
@@ -262,6 +272,8 @@ LibraryView shows both with indicators:
 | `validate-token` | invoke | Validate API token against smotret-anime.ru |
 | `search-anime` | invoke | Search anime by query |
 | `get-anime` | invoke | Fetch anime details by ID |
+| `get-anime-cache` | invoke | Read cached AnimeDetail for fast first paint (returns null if missing or older than 24h) |
+| `set-anime-cache` | invoke | Write AnimeDetail to cache (no-op if anime is neither starred nor downloaded) |
 | `get-episode` | invoke | Fetch episode translations |
 | `probe-embed-quality` | invoke | Probe embed API for actual stream height |
 | `report-quality-mismatch` | invoke | Report a detected quality mismatch (stored in memory) |
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "anime-dl-app",
-  "version": "3.20.2",
+  "version": "3.20.3",
   "description": "Anime episode downloader",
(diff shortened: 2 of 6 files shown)


akvasha:
## Follow-up: Shikimori section is still slow

This PR makes the AnimeDetailView render instantly from cache, but the Shikimori panel still blocks on three live API calls — `get-rate`, `get-friends-rates`, and `get-related`. Suggested follow-up (defer to a separate PR):

1. **`get-rate` — cache-first from existing `shikimoriUserRates` store.**
   - The user's own rates are already mirrored locally; render from the store immediately, then revalidate via API.
   - No new cache surface needed; just flip the read order.

2. **`get-friends-rates` — new `shikimoriFriendsRatesCache` with 30-min TTL.**
   - Friends' ratings change relatively often, so a short TTL is fine.
   - Emit `shikimori:friends-rates-updated` after background refresh so the UI can re-render without a reload.

3. **`get-related` — new `shikimoriRelatedCache` with 24h TTL.**
   - Related-anime lists are essentially static; long TTL is safe.
   - Emit `shikimori:related-updated` on refresh.

Same cache-first + background-refresh + generation-guard pattern as this PR — should be a small, mechanical follow-up.

mikkerlo (approved):
LGTM after manual testing

PR #74, opened 2026-05-06
fix(calendar): show airing episode, not user's next-up

### LINKED ISSUE #73: Airing Calendar shows user's next episode instead of airing episode

## Motivation
Currently, the airing calendar displays the next episode the user needs to watch (watched episodes + 1) instead of the actual episode that is scheduled to air. This confuses users who expect a calendar to show upcoming releases rather than their personal watch progress.

## Approach & Architecture
The Shikimori API `/api/calendar` endpoint returns a `next_episode` field which contains the exact episode number that is scheduled to air at `next_episode_at`.
We need to update the `ShikiCalendarEntry` interface to include this field and modify the backend IPC handler to map this value to `episodeInt` instead of calculating it from the users local watched episodes.

The UI in `CalendarView.vue` currently displays "Ep X" using `entry.episodeInt`. No UI changes are strictly required as long as `episodeInt` correctly represents the airing episode.

## Implementation Plan

   - [ ] Update `ShikiCalendarEntry` in `src/main/shikimori.ts` to include `next_episode: number`.
   - [ ] Update the `shikimori:get-calendar` handler in `src/main/index.ts` to map `c.next_episode` to the `episodeInt` field of the `CalendarEntry` being returned. Convert it to a string.

## Files to Touch
- **Modify:** `src/main/shikimori.ts`

## Testing Strategy
- Open the Airing Calendar.
- Verify that for an anime where you are behind on episodes (e.g., watched 1, airing episode is 5), the calendar displays "Ep 5" instead of "Ep 2".
- Verify that the calendar still loads correctly.

## Risks & Edge Cases
- Ensure fallback behavior in case `next_episode` is not provided by the API.

## Out of Scope
- Adding UI to show personal watch progress alongside the airing episode.

akvasha:
## Summary

The Airing Calendar was labelling each card with the user's personal next-up episode (`rate.episodes + 1`) instead of the episode actually scheduled to air at `next_episode_at`. For viewers behind on an ongoing show, this misrepresented what was on the schedule — e.g. a user who has watched 1 of 5 aired episodes saw "Ep 2" on a slot whose airing episode was 5.

This PR pivots the calendar to show what's airing, decoupling it from the viewer's progress.

## Changes

- **`src/main/shikimori.ts`** — added `next_episode: number | null` to `ShikiCalendarEntry`. Shikimori's `/api/calendar` already returns this alongside `next_episode_at`; we just weren't deserializing it.
- **`src/main/index.ts`** — `shikimori:get-calendar` now sets `episodeInt` from `c.next_episode ?? (c.anime.episodes_aired + 1)`. The fallback keeps airing semantics if the API omits the field. The per-user `episodesByMal` map and the `watchedEps` lookup were the only consumers of personal progress in this handler and have been removed.
- **`DESIGN.md`** — rewrote the line in *Airing Calendar → Data flow* that documented `episodeInt = rate.episodes + 1` so the doc matches the new semantics.
- **`package.json`** — patch bump to `3.20.3`.

The IPC return shape (`CalendarEntry`) is unchanged, so no preload / `types.d.ts` / renderer edits were needed.

## Behavior change

Before: card showed the user's next-up episode for that show.
After: card shows the episode Shikimori reports as next-airing.

For users who are caught up, this is identical to the old behavior. For users who are behind, the calendar now reflects the schedule rather than their backlog. There's no UI signal of "you're behind by N" — that was discussed but is intentionally out of scope here; track separately if wanted.

## Cache note

`calendarCache` is in-memory only and short-TTL. The new response is user-independent, but the cache key didn't depend on user identity to begin with, so no invalidation logic changes.

## Test plan

- [x] \`npm run typecheck\` passes.
- [x] Open the Airing Calendar with a tracked, currently-airing show where \`watched < episodes_aired\`. Card shows the airing episode, not \`watched + 1\`.
- [x] Click the topbar refresh (forces \`force=true\`, bypasses the in-memory cache). Entries still render.
- [x] Calendar still loads on the "planned" path (a tracked show you haven't started).

Fixes #73


mikkerlo (approved):
The fix correctly pulls the `next_episode` field from the Shikimori API and gracefully falls back to `episodes_aired + 1` if it's missing. The cleanup of unused variables and documentation updates are also spot on.

LGTM after manual testing

PR #75, opened 2026-05-06
feat: manual OP/ED chapter injection into merged MKVs

### LINKED ISSUE #64: Add OP/ED skip points as chapters to merged MKV files

## Motivation
Currently, the application detects openings and endings (OP/ED) for downloaded episodes, but this information is only available within the built-in player. If a user opens the MKV file in an external player (like VLC or mpv), they lose the ability to easily skip these sections. Instead of injecting chapters automatically during the initial merge (which can be fragile or unwanted), we need a manual way to extract detected OP/ED timings and save them as chapters into existing MKVs.

## Approach & Architecture
We will replace the planned auto-merge injection with a manual "Analyze and save chapters" action per anime.
When triggered for an anime with downloaded episodes, the app will:
1. Ensure OP/ED skip points are detected for the episodes.
2. For each downloaded episode, generate an FFMETADATA1 file defining chapters (Intro, OP, Episode, ED, Outro) using precision milliseconds.
3. Pass this temporary metadata file to `ffmpeg` and map specifically for chapters (`-map_chapters <idx>`), ensuring we don't overwrite global metadata, and rewrite the MKV file to include these chapters.
4. Update the `skipFingerprintCache` to reflect the newly written file's mtime and size.

- Add a new manual "Analyze and save chapters" button on the anime details view (or downloads view).
- The button should ONLY be visible/enabled if the anime has at least 3 downloaded episodes.
- Remove the originally planned `enableChapterMarking` toggle from Settings since the process is now manual.
- Show progress/loading state while the chapters are being saved to MKVs.

## Implementation Plan

   - [ ] `src/main/skip-detector.ts`: Implement `formatChaptersMetadata(durationSec, op, ed)` returning an FFMETADATA1 string (or null if no chapters should be emitted).
     - Use `TIMEBASE=1/1000` and emit milliseconds (`Math.round(startSec * 1000)`).
     - Chapter titles: `Intro`, `OP`, `Episode`, `ED`, `Outro`.
     - Implement FFMETADATA escaping for chapter titles (`=`, `;`, `#`, `\`, `\n`).
     - Drop degenerate segments shorter than 2s.
     - Validate OP/ED ordering to drop or clamp inverted ranges (e.g. `op.endSec > ed.startSec`).
   - [ ] `src/main/download-manager.ts`:
     - Update `runFfmpeg` signature to accept an options bag: `runFfmpeg(opts: { videoPath, subtitlePath?, outputPath, codec, subMeta?, chaptersMeta?, onPercent? })`.
     - Thread the correct input index for `-map_chapters` depending on whether a subtitle input is present.
   - [ ] `src/main/index.ts`:
     - Create a new IPC handler `download:inject-chapters` that takes an `animeId`.
     - For each downloaded episode of `animeId`, conditionally generate chapters using `formatChaptersMetadata` and pass them to `runFfmpeg`.
     - Update the `skipFingerprintCache` key to the new mtime/size after a successful rewrite so the cache remains valid.

   - *Note: Remember to update both `preload/index.ts` and `preload/types.d.ts` for any new IPC channels.*
   - [ ] Add `injectChapters(animeId: number): Promise<void>` to the IPC API.

   - [ ] `src/renderer/src/components/AnimeDetailView.vue` (or equivalent): Add an "Analyze and save chapters" button.
   - [ ] Make the button visible/enabled ONLY if there are at least 3 downloaded episodes for the anime.
   - [ ] Handle loading state while the IPC call is running.

## Files to Touch
- **New:** None
- **Modify:** `src/main/skip-detector.ts`
- **Modify:** `src/main/download-manager.ts`
- **Modify:** `src/renderer/src/components/AnimeDetailView.vue` (or similar view)
- **Documentation:** `DESIGN.md` *(Document the new chapter injection architecture and IPC)*

## Testing Strategy
- **Idempotency:** Re-run `download:inject-chapters` on a file that already has chapters injected by this same code path; verify with `ffprobe -show_chapters` that chapters are not duplicated.
- **Condition:** Verify the button only appears when >= 3 downloaded episodes are present.
- **Edge cases:**
  - Episode where OP starts at second 0 (should not create an `Intro` chapter).
  - Episode with no OP and no ED (should skip chapter generation).
  - Episode shorter than 2 × OP duration (music videos/specials).

## Risks & Edge Cases
- **Metadata Overwrite:** Using `-map_metadata` without care can overwrite global tags. We must use `-map_chapters <idx>` and leave global metadata alone (or explicitly `-map_metadata 0 -map_chapters <idx>`).
- **Cache Invalidation:** MKV rewrite invalidates `skipFingerprintCache`. We must manually update the cache key with the new file size and mtime.
- **Garbage Output:** Overlapping OP/ED times could cause external players to fail parsing chapters. We must validate and clamp ranges.

## Out of Scope
- Modifying the in-app player's behavior based on MKV chapters (it continues to use the local/remote detection database).
- Generating chapters for anything other than OP/ED/Episode (e.g., scene changes).

akvasha:
## Summary
- Adds a **Save chapters to MKV** button in the Skip Detection panel of `AnimeDetailView`. Visible only when at least 3 merged `.mkv` episodes are on disk.
- Rewrites each merged MKV in place with `Intro` / `OP` / `Episode` / `ED` / `Outro` chapters derived from cached `skipDetections`. If no signature exists yet, runs a fresh analysis first.
- Deliberately **manual** — no auto-injection during the original merge, so a bad detection can never silently corrupt new files.

## Changes
- `src/main/skip-detector.ts` — new `formatChaptersMetadata(durationSec, op, ed)` returning an FFMETADATA1 string. `TIMEBASE=1/1000`, ms-precision starts/ends, drops sub-2s segments, drops ED on inverted ranges (OP end past ED start), escapes `= ; # \\ \n` in titles. Returns `null` when nothing useful would be written.
- `src/main/download-manager.ts` — refactored `runFfmpeg` to an options bag and made it `public`. New `chaptersMeta` field writes a temp metadata file and adds `-map_chapters <idx>` (idx = subtitle ? 2 : 1) so global metadata stays untouched. Cleaned up on both success and error paths.
- `src/main/index.ts` — new IPC `download:inject-chapters(animeId, episodes)`. Filters to existing `.mkv` files, requires ≥3, lazily kicks off `runSkipAnalysisInternal` if no detections, writes to `.<base>.chapters.tmp.mkv` then renames over the original, remaps the `skipFingerprintCache` key to the new size/mtime. Broadcasts `chapter-inject:progress` per episode.
- `src/preload/index.ts` + `src/preload/types.d.ts` — `injectChapters(animeId, episodes)` plus on/off helpers for `chapter-inject:progress`.
- `src/renderer/src/components/AnimeDetailView.vue` — button, progress text, error & summary lines. Reuses the renderer's already-computed `skipEpisodeInputs` so the button-visibility check and the IPC see the same set of files.
- `DESIGN.md` — documents the IPC and the rollout-phase entry.
- `package.json` — `3.20.3 → 3.21.0`.

## Behavior
- Idempotent: re-running on a file that already has these chapters does not duplicate them — ffmpeg replaces the chapter list when `-map_chapters` is mapped.
- Cache-friendly: the audio fingerprint doesn't change when chapters are rewritten, so the cache entry is moved to the new `(size, mtime)` key rather than discarded.
- If a per-episode detection lacks usable OP and ED, that episode is skipped (no rewrite, no error).

## Test plan
- [x] `npm run typecheck` passes.
- [x] Verified end-to-end on Classroom of the Elite S4 (5 merged MKVs) — `ffprobe -show_chapters` shows `Intro` / `OP` / `Episode` / `ED` on the rewritten files. Episode 01 had no OP/ED detection and was skipped (expected).
- [x] Confirmed chapters appear and seekable in mpv.net (`PgUp`/`PgDn`).
- [ ] Re-click "Save chapters to MKV" on the same anime — chapter count stays the same (no duplicates).
- [ ] Built-in player still renders OP/ED bands on rewritten files (cache key remap).
- [ ] Edge: verify nothing breaks when only OP is detected (no ED) or vice versa.

Fixes #64


src/main/index.ts
```diff
@@ -2695,6 +2695,77 @@ function registerIpcHandlers(): void {
...
+        const cached = cache[oldKey]
+
+        await downloadManager.runFfmpeg({
+          ffmpegPath,
+          ffprobePath,
+          videoPath: ep.filePath,
+          outputPath: tmpOut,
+          codec: 'copy',
+          chaptersMeta
+        })
```
mikkerlo: Reusing `downloadManager.runFfmpeg` introduces a concurrency bug. `runFfmpeg` manages the state for the download queue (it assigns `this.activeFfmpegCmd` and clears `this.activeMergeTranslationId`). If chapter injection runs concurrently with an episode merge, it will clobber these pointers, breaking `cancelMerge()`. Additionally, when no subtitle path is provided, `runFfmpeg` only specifies `-c:v copy` and `-c:a copy`, which means the existing subtitle track might be re-encoded, and attachments like fonts will be dropped since `-map 0` isn't used.

Instead, please spawn a dedicated, stateless `Ffmpeg` pipeline here. To safely copy all original tracks and attachments perfectly, use `-map 0` and `-c copy`:

```suggestion
        const chaptersMetaFile = path.join(os.tmpdir(), `anime-dl-chapters-${process.pid}-${Date.now()}.txt`)
        fs.writeFileSync(chaptersMetaFile, chaptersMeta, 'utf8')

        try {
          await new Promise<void>((resolve, reject) => {
            Ffmpeg.setFfmpegPath(ffmpegPath)
            Ffmpeg(ep.filePath)
              .input(chaptersMetaFile)
              .outputOptions('-y')
              .outputOptions(['-map', '0', '-map_chapters', '1', '-c', 'copy'])
              .output(tmpOut)
              .on('end', resolve)
              .on('error', reject)
              .run()
          })
        } finally {
          try { fs.unlinkSync(chaptersMetaFile) } catch { /* ignore */ }
        }
```

mikkerlo (changes requested):
The PR implements the core functionality well and perfectly follows the specifications from the issue plan. However, there is a critical concurrency bug caused by reusing a stateful method from `DownloadManager`, and a minor risk of stream loss during the remux. I have requested changes with a line-level comment detailing the fix. Please address this issue and I will approve the PR.

akvasha:
1. **Concurrency + missing `-map 0`** — fixed in f2f9467. The IPC handler now spawns a dedicated stateless `Ffmpeg(...)` pipeline directly (no `downloadManager.runFfmpeg` reuse), so it can't clobber `activeFfmpegCmd` / `activeMergeTranslationId` if an inject overlaps a merge. Switched the output options to `-map 0 -map_chapters 1 -c copy` so subtitle tracks, fonts, and other attachments are passthrough-copied as-is. Reverted `runFfmpeg` to its original private signature and dropped the now-unused `chaptersMeta` plumbing — `runFfmpeg` is back to "merge only".

src/main/index.ts
```diff
@@ -2695,6 +2695,84 @@ function registerIpcHandlers(): void {
...
+            .run()
+        })
+        fs.renameSync(tmpOut, ep.filePath)
+
+        if (cached) {
+          const newStat = fs.statSync(ep.filePath)
+          const newKey = fingerprintCacheKey(animeId, ep.episodeInt, newStat.size, newStat.mtimeMs)
+          delete cache[oldKey]
+          cache[newKey] = { ...cached, fileSize: newStat.size, fileMtimeMs: newStat.mtimeMs }
+        }
```
mikkerlo, line 2761:
```suggestion
          store.set('skipFingerprintCache', cache)
        }
```

nit: Since MKV remuxing can take a while for a large batch, persisting the cache *inside* the loop right after each successful rename is safer. If the user quits the app mid-way, `sweepSkipFingerprintCache()` on the next launch will evict the entries for the already-remuxed files because their mtimes changed but the cache wasn't persisted yet.

src/main/index.ts
```diff
@@ -2695,6 +2695,84 @@ function registerIpcHandlers(): void {
...
+        failed++
+        try { fs.unlinkSync(tmpOut) } catch { /* ignore */ }
+        const msg = err instanceof Error ? err.message : String(err)
+        console.error(`[chapter-inject] failed for ${ep.episodeLabel}: ${msg}`)
+      } finally {
+        try { fs.unlinkSync(chaptersMetaFile) } catch { /* ignore */ }
+      }
+    }
+
+    store.set('skipFingerprintCache', cache)
```
mikkerlo:
```suggestion
```

(Remove this, as it's now handled safely inside the loop above)

mikkerlo (approved):
LGTM after manual testing

akvasha:
1. **Persist `skipFingerprintCache` per-episode (L2760)** — fixed in 9cb508b. The `store.set` now fires inside the loop right after each successful rename, so a mid-batch quit leaves the cache consistent with the already-rewritten files instead of forcing `sweepSkipFingerprintCache()` to evict them on next launch.
2. **Drop trailing `store.set` (L2772)** — fixed in the same commit. Removed since per-episode persistence makes it redundant.

PR #79, opened 2026-05-07
feat: auto-download newly aired episodes for subscribed shows (#76)

### LINKED ISSUE #76: Auto-download newly aired episodes for subscribed shows

## Motivation

Today, watching a currently-airing show is a manual loop: episode airs → check `CalendarView` → open `AnimeDetailView` → wait for the new episode row to populate → click Download. For users tracking 5–10 simultaneous airing shows, this means returning to the app every few days and repeating the same chore for each one.

We already have all the building blocks: `shikimori:get-calendar` knows which tracked shows have new episodes (and when), `DownloadManager` knows how to queue, and `animePrefs` remembers each show's preferred translation. Glue them together so that opting a show into "auto-download" makes new episodes show up on disk without any further interaction.

## Approach & Architecture

A new main-process worker (`src/main/auto-downloader.ts`) that:

1. Reads a per-show subscription map from a new `autoDownloadSubscriptions` electron-store key.
2. On each tick, walks every subscribed show, fetches the latest aired episode count (via existing `shikimori:get-calendar` cache + `shikimoriUserRates` cache — Shikimori `episodes_aired` is the source of truth), and identifies forward-only newly-aired episodes (those with `episodeInt > subscription.lastSeenEpisodesAired`).
3. For each newly-aired episode, resolves a `DownloadRequest` exactly like the manual path does: pick translation via priority chain (`subscription.translationType/author` → `animePrefs[animeId]` → global defaults), call `getEpisode(animeId)` to find the matching translation row, probe its real quality with `probe-embed-quality`, build a `DownloadRequest` and feed it to the existing `download:enqueue` handler.
4. Dedupes against `downloadedEpisodes[animeId:episodeInt]` (already-downloaded), the live `download-manager` queue (already-queued), and an in-memory "this tick already enqueued" set to prevent overlapping ticks from double-queueing.
5. Updates `subscription.lastSeenEpisodesAired` after a successful enqueue so subsequent ticks don't re-queue the same episode even if the download later fails.

**Tick triggers** (in OR fashion, all gated by a single 60s reentrancy lock):
- App start (~30s after `app.whenReady` fires, after Shikimori rates load)
- Periodic timer every 15 minutes
- Reactive: on `shikimori:rates-refreshed` (the user just refreshed their watchlist) and on `shikimori:get-calendar` cache invalidation
- Manual: a "Run now" button in `Settings > General > Auto-download`

**Rate-limit / safety guards**:
- `MAX_ENQUEUES_PER_TICK = 10` — a hard cap on episodes queued in a single worker run, so a misconfigured `lastSeenEpisodesAired` (or a show that suddenly returns 200 episodes_aired) can't dump 100s of items into the queue.
- The forward-only design (subscription stamps `lastSeenEpisodesAired = current episodes_aired` at subscribe time) means newly-subscribed shows never backfill — only episodes that air after the subscription click are queued.
- Per-show subscription disable also pauses any in-flight queued items for that show? **No** — disabling stops future enqueues; existing queue items keep flowing. (Symmetric with manual downloads.)

**Subscription record shape** (`autoDownloadSubscriptions[animeId]`):
```ts
interface AutoDownloadSubscription {
  animeId: number
  malId: number
  subscribedAt: number          // epoch ms
  lastSeenEpisodesAired: number // bumped after each successful enqueue
  translationType?: string      // optional override; falls back to animePrefs → global
  author?: string               // optional override; falls back to animePrefs → global
}
```

## UI / UX Considerations

- **AnimeDetailView**: a new subscribe toggle button next to the existing star/library button — a small pill labeled "Auto-download" with a clock/repeat icon. Tooltip explains the forward-only behavior ("New episodes will be downloaded automatically as they air."). Visible only for ongoing shows (i.e. when the show has `episodes_aired < total_episodes` or is `ongoing`/`anons` on Shikimori). Greyed out (with explanation) when the user is not connected to Shikimori — the calendar pipeline depends on it.
- **CalendarView**: subscribed shows get a small ↻ chip on their card so the user can see at a glance which entries will auto-download.
- **Settings > General > Auto-download** (new section):
  - Master toggle: "Auto-download new episodes for subscribed shows" (default off the first time the user opens the app post-feature; on once they subscribe to anything — see migration note in Risks).
  - "Run now" button (calls `auto-dl:trigger`).
  - Read-only counter: "X shows subscribed", clicking opens a small modal with the list (anime name + last-checked timestamp + per-show unsubscribe button).
- **Notifications**: piggyback on the existing `notificationMode` setting — the system notification that fires when a download completes already labels the show, so no new notification UX is needed. We add an in-flight broadcast `auto-dl:enqueued` so the user sees a brief toast when a subscription kicks in (only when the app is foregrounded).

## Implementation Plan

   - [ ] New file `src/main/auto-downloader.ts` exporting `runAutoDownloadTick(opts: { reason: 'startup'|'timer'|'rates-refreshed'|'calendar-refresh'|'manual' })`. Implements the tick logic, single 60s reentrancy lock, `MAX_ENQUEUES_PER_TICK` cap, dedupe set.
   - [ ] Translation/quality resolver helper: given `(animeId, malId, episodeInt, subscription)`, returns a `DownloadRequest` or `null` (and a reason: `no-translation` / `not-on-smotret` / `already-downloaded` / `already-queued` / `embed-failed`).
   - [ ] Wire up triggers in `src/main/index.ts`: `setInterval(15 * 60 * 1000)`, post-`whenReady` startup tick, listener bumps on the IPC paths that already touch `shikimoriUserRates`/calendar cache.
   - [ ] New IPC handlers: `auto-dl:get-subscription` (animeId), `auto-dl:set-subscription` (animeId, enabled, optional overrides), `auto-dl:list-subscriptions`, `auto-dl:trigger`, `auto-dl:get-status` (last-tick timestamp + result counters).
   - [ ] New broadcast: `auto-dl:tick-result` ({ enqueued, skipped, errors, ranAt }) so the renderer can refresh the status panel without polling.
   - [ ] New broadcast: `auto-dl:enqueued` ({ animeId, episodeInt, animeName }) for the foreground toast.
   - [ ] New store key + default in `electron-store` schema.
   - *Note: update both `src/preload/index.ts` and `src/preload/types.d.ts` for every new IPC channel above.*
   - [ ] Expose `autoDlGetSubscription`, `autoDlSetSubscription`, `autoDlListSubscriptions`, `autoDlTrigger`, `autoDlGetStatus`.
   - [ ] Expose listener helpers `onAutoDlTickResult`, `onAutoDlEnqueued`.
   - [ ] `AnimeDetailView.vue`: add subscribe toggle button, hydrate state from `autoDlGetSubscription`, save with `autoDlSetSubscription`. Hide for shows without a `malId` (we can't resolve airing data) and for shows already finished. Show a small "Will catch new episodes from EpN onward" hint under the toggle, where N = current `episodes_aired + 1`.
   - [ ] `CalendarView.vue`: query `autoDlListSubscriptions` once per load; render the ↻ chip on cards whose `animeId` is subscribed.
   - [ ] `SettingsView.vue`: new "Auto-download" section under General — global toggle, "Run now", counter + modal listing subscriptions.
   - [ ] Top-level toast handler in `App.vue` (or wherever existing toasts live) for `auto-dl:enqueued`.

## Files to Touch

- **New:** `src/main/auto-downloader.ts`
- **Modify:** `src/main/index.ts` (IPC handlers + tick wiring + store schema entry)
- **Modify:** `src/renderer/src/components/CalendarView.vue`
- **Modify:** `src/renderer/src/App.vue` (toast hookup, if needed)
- **Documentation:** `DESIGN.md` (new IPC channels, new store keys, new auto-downloader section)

## Testing Strategy

- **Forward-only stamp**: subscribe to an ongoing show with `episodes_aired = 5`. Wait for next episode. Verify only ep 6 is queued, never eps 1–5.
- **Translation resolution**: subscribe to a show after manually downloading ep 5 with `voiceRu/AniDub`. Verify ep 6 picks up the same translation, not the global default.
- **Override path**: set per-subscription `translationType = subRu`. Verify ep 6 gets `subRu` even if `animePrefs` says otherwise.
- **Dedupe**: manually queue ep 6, then run `auto-dl:trigger`. Verify the worker skips ep 6.
- **Already downloaded**: have ep 6 on disk, `episodes_aired` jumps to 7. Verify only ep 7 is queued.
- **Rate cap**: simulate a subscription where `lastSeenEpisodesAired = 0` and `episodes_aired = 50`. Verify only `MAX_ENQUEUES_PER_TICK` items are queued in one tick, and the next tick picks up where it left off.
- **No Shikimori**: log out; verify the subscribe button is disabled with a clear "Connect to Shikimori" message and ticks become no-ops without errors.
- **Show not on smotret-anime**: subscribe to an MAL-only entry (no `malIdMap` resolution); verify the worker logs "not-on-smotret" and doesn't retry every tick (de-prioritize via a backoff field on the subscription).
- **Reentrancy**: kick `auto-dl:trigger` 5 times rapidly. Verify only one tick runs and the rest are coalesced.
- **Cross-platform**: Windows/Mac/Linux paths shouldn't matter (worker is renderer-process-agnostic), but verify notifications fire on each.

## Risks & Edge Cases

- **Embed URL freshness**: smotret-anime stream URLs expire. The auto-downloader runs the embed probe inline at enqueue time (same as manual path), so this is fine — but if `processQueue` defers a queued item by hours, the existing 416-retry path must handle the URL having gone stale. Verify on a delayed queue.
- **Calendar API drift**: if Shikimori reports `episodes_aired` jumping non-monotonically (rare, but possible on schedule corrections), the worker might re-queue an episode the user already grabbed. Dedupe against `downloadedEpisodes` covers this.
- **Translation availability lag**: a newly-aired episode may show up in the Shikimori calendar before any translation is available on smotret-anime. The resolver returns `no-translation` and the tick stamps a *separate* `lastTranslationCheckAt` field on the subscription so it backs off (e.g. retry once per tick for 24h, then daily) instead of hammering `getEpisode` every 15 min. **Note**: this means the `lastSeenEpisodesAired` stamp is only bumped *after* a successful enqueue or a successful "already downloaded" detection — never on `no-translation`.
- **User unsubscribes mid-tick**: handled because we read the subscription set fresh inside the lock; an unsubscribe between ticks just removes the show from the next pass.
- **Quota / disk-full**: existing download-manager error path is reused; failed downloads surface in DownloadsView as today.
- **Double-firing on `rates-refreshed`**: the 60s reentrancy lock + reason aggregation prevents two ticks from running back-to-back even if multiple triggers fire within the lock window.
- **Concurrent app instances**: electron-store is process-local; if the user runs two app instances, both could enqueue the same episode. Out of scope — single-instance lock already exists for the main app, but call it out.
- **Migration**: the master Settings toggle should default to **on** for users who explicitly subscribe to a show — i.e. the first `auto-dl:set-subscription` call also flips the master toggle on (with a one-shot toast explaining it). This avoids the confusing case where the user subscribes but the global gate silently blocks anything from happening.

## Out of Scope

- **Backfill of past missed episodes** — explicitly forward-only per the answered scoping question. Could be a follow-up issue ("Catch up since last watched") with its own UX warning.
- **Auto-download for `Planned` shows** — only `Watching`/`Rewatching` shows trigger auto-downloads; planned shows still appear in the calendar but don't subscribe-by-default. Subscribing manually to a planned show works but the user has to opt in show-by-show.
- **Per-show schedule overrides** ("only download Tuesday episodes") — niche; handle if requested.
- **Auto-merge / auto-move-to-cold gating** — these continue to obey their existing global settings; no new wiring.
- **Push notification when subscription falls behind** (e.g. 3 days of `no-translation`) — could be a follow-up if users hit this regularly.
- **Smart quality picker that depends on free disk space** — out of scope; uses the existing global default quality.
- **Background ticks while app is closed** — Electron only runs while open; users who quit the app won't see auto-downloads until next launch (the startup tick will catch up at that point, bounded by `MAX_ENQUEUES_PER_TICK`).

akvasha:
## Summary

Closes #76. Subscribe to airing shows from `AnimeDetailView` and the app queues new episodes automatically — forward-only, no backfill, gated by Shikimori login + valid MAL ID.

## Changes

- **New** `src/main/auto-downloader.ts` — worker with 60s reentrancy lock, `MAX_ENQUEUES_PER_TICK=10`, in-tick dedup
- **Triggers** in `src/main/index.ts`: `+30s` after `whenReady`, `setInterval(15min)`, post-`shikimori:rates-refreshed`, manual via `auto-dl:trigger` IPC
- **Inline detail refresh** — when cached `shikimoriAnimeDetails[malId]` is older than 6h, the tick refetches via `shikimori.getAnimeDetails` so weekly-airing shows don't go stale between the existing pre-fetch worker's 30-day cadence
- **Translation resolver** — prefers most-recent `downloadedEpisodes` entry for the show (`type` + `author`), falls back to global `translationType` default
- **Dedup** vs `downloadedEpisodes` (new + legacy key formats), live `downloadManager.getEpisodeGroups()`, and per-tick set
- **Forward-only stamping** — at subscribe time, `lastEnqueuedEpisodeInt = current episodes_aired`; bumped on successful enqueue or already-downloaded/already-queued; never on `no-translation`
- **IPC** — 7 handlers (`auto-dl:get-subscription`, `auto-dl:set-subscription`, `auto-dl:list-subscriptions`, `auto-dl:trigger`, `auto-dl:get-status`, `auto-dl:get-enabled`, `auto-dl:set-enabled`) and 2 broadcasts (`auto-dl:tick-result`, `auto-dl:enqueued`)
- **Store keys** — `autoDownloadSubscriptions` (per-show map), `autoDownloadEnabled` (master toggle, default on)
- **UI** —
  - `AnimeDetailView.vue` — Auto-download pill next to library star (hidden for finished shows or when not connected to Shikimori)
  - `CalendarView.vue` — ↻ Auto-DL chip on subscribed cards
  - `SettingsView.vue` — General → Auto-download section with master toggle, **Run now**, last-run counters, expandable subscription list with per-show unsubscribe
- **`DESIGN.md`** — IPC table, store schema, and a new Auto-download subsection under Shikimori
- **Version bump** 3.21.0 → 3.22.0

## Behavior notes

- Master toggle defaults to `true`; turning it off makes ticks no-ops without dropping subscriptions
- Manual "Run now" that hits the 60s reentrancy lock returns silently (no broadcast); SettingsView reads `autoDlLastResult` only from broadcasts, so blocked clicks don't overwrite real counters
- `MAX_ENQUEUES_PER_TICK=10` is a hard cap — if a show jumps from 0→50 episodes_aired, only 10 enqueue per run, the rest pick up next tick
- Subscriptions are gated on `myAnimeListId` + Shikimori login (the calendar pipeline depends on Shikimori) + `status !== 'released'`

## Test plan

- [ ] Subscribe to an ongoing show with episodes_aired = N → click Run now → verify nothing enqueues (forward-only stamp)
- [ ] Manually decrement lastEnqueuedEpisodeInt in ~/.config/anime-dl-app/config.json → restart app → click Run now → verify ep N enqueues with the right translation
- [ ] Manually queue an episode ahead → click Run now → verify dedup skips it (already-queued)
- [ ] Stale cache: subscribe to a show whose shikimoriAnimeDetails cache is >6h old → click Run now → verify the tick refetches details inline and queues newly-aired episodes
- [ ] Subscribe to a finished show → verify the pill is hidden
- [ ] Log out of Shikimori → verify the pill is disabled and ticks become no-ops
- [ ] Stamp lastEnqueuedEpisodeInt 50 episodes behind → verify only 10 enqueue per tick
- [ ] Disable master toggle → verify Run now broadcasts enqueued: 0
- [ ] Cross-platform smoke: Windows/macOS/Linux (no platform-specific code paths)

### DIFF
--- a/src/main/auto-downloader.ts
+++ b/src/main/auto-downloader.ts
@@ -102,11 +102,11 @@ export function getSubscription(animeId: number): AutoDownloadSubscription | nul
   return map[String(animeId)] ?? null
 }
 
-export function setSubscription(
+export async function setSubscription(
   animeId: number,
   enabled: boolean,
   meta?: { malId: number; animeName: string }
-): AutoDownloadSubscription | null {
+): Promise<AutoDownloadSubscription | null> {
   if (!deps) return null
   const map = { ...(deps.store.get('autoDownloadSubscriptions') as Record<string, AutoDownloadSubscription>) }
   const key = String(animeId)
@@ -123,9 +123,22 @@ export function setSubscription(
     return existing
   }
   // Stamp lastEnqueuedEpisodeInt to the current episodes_aired so we never backfill.
+  // If the cache is missing or stale, inline-refresh first — otherwise an empty cache
+  // would leave the stamp at 0 and the next tick would download the entire show.
   const cache = deps.store.get('shikimoriAnimeDetails') as Record<string, ShikiAnimeDetailsCacheEntry>
-  const cacheEntry = cache[String(meta.malId)]
+  let cacheEntry = cache[String(meta.malId)]
+  const cacheAge = cacheEntry ? Date.now() - cacheEntry.fetchedAt : Infinity
+  if (!cacheEntry || cacheAge > DETAILS_FRESHNESS_MS) {
+    const fresh = await deps.refreshShikimoriDetails(meta.malId)
+    if (fresh) {
+      cacheEntry = { details: fresh, fetchedAt: Date.now() }
+    }
+  }
   const aired = cacheEntry?.details?.episodes_aired ?? 0
+  if (aired <= 0) {
+    // Refuse rather than silently arming a backfill. Caller surfaces this to the user.
+    return null
+  }
   const sub: AutoDownloadSubscription = {
     animeId,
     malId: meta.malId,
@@ -408,6 +421,6 @@ export async function runAutoDownloadTick(reason: AutoDlReason): Promise<AutoDlT
     return result
   } finally {
     tickRunning = false
-    lockReleaseAt = 0
+    // Keep lockReleaseAt as the cooldown deadline so manual triggers actually wait REENTRANCY_LOCK_MS.
   }
 }
--- a/src/main/auto-downloader.ts
+++ b/src/main/auto-downloader.ts
@@ -1,0 +1,413 @@
+import type Store from 'electron-store'
+import type { SmotretApi, EpisodeDetail, Translation } from './smotret-api'
+import type { DownloadManager, DownloadRequest } from './download-manager'
+import type { AutoDownloadSubscription } from './index'
+import type * as shikimori from './shikimori'
+
+export type AutoDlReason = 'startup' | 'timer' | 'rates-refreshed' | 'manual'
+
+export interface AutoDlOutcome {
+  animeId: number
+  animeName: string
+  episodeInt: string
+  outcome:
+    | 'enqueued'
+    | 'no-translation'
+    | 'no-episode'
+    | 'already-downloaded'
+    | 'already-queued'
+    | 'embed-failed'
+    | 'no-episodes-aired'
+    | 'cap-reached'
+    | 'error'
+  message?: string
+}
+
+export interface AutoDlTickResult {
+  ranAt: number
+  reason: AutoDlReason
+  enqueued: number
+  skipped: number
+  errors: number
+  details: AutoDlOutcome[]
+}
+
+interface DownloadedEpisodeMeta {
+  translationType: string
+  author: string
+  quality: number
+  translationId: number
+}
+
+interface ShikiAnimeDetailsCacheEntry {
+  details: shikimori.ShikiAnimeDetails
+  fetchedAt: number
+}
+
+interface AutoDownloaderDeps {
+  store: Store<Record<string, unknown>>
+  smotretApi: SmotretApi
+  downloadManager: DownloadManager
+  broadcast: (channel: string, ...args: unknown[]) => void
+  isShikimoriLoggedIn: () => boolean
+  refreshShikimoriDetails: (malId: number) => Promise<shikimori.ShikiAnimeDetails | null>
+}
+
+const MAX_ENQUEUES_PER_TICK = 10
+const REENTRANCY_LOCK_MS = 60_000
+const TICK_INTERVAL_MS = 15 * 60 * 1000
+const DETAILS_FRESHNESS_MS = 6 * 60 * 60 * 1000
+
+let deps: AutoDownloaderDeps | null = null
+let lastTickResult: AutoDlTickResult | null = null
+let tickRunning = false
+let lockReleaseAt = 0
+let timerHandle: ReturnType<typeof setInterval> | null = null
+
+export function initAutoDownloader(d: AutoDownloaderDeps): void {
+  deps = d
+}
+
+export function startAutoDownloaderTimer(): void {
+  if (timerHandle) return
+  timerHandle = setInterval(() => {
+    void runAutoDownloadTick('timer').catch((err) => {
+      console.warn('[auto-dl] timer tick failed:', err)
+    })
+  }, TICK_INTERVAL_MS)
+}
+
+export function getAutoDownloaderStatus(): {
+  lastResult: AutoDlTickResult | null
+  locked: boolean
+  enabled: boolean
+} {
+  const enabled = deps ? Boolean(deps.store.get('autoDownloadEnabled')) : false
+  return {
+    lastResult: lastTickResult,
+    locked: tickRunning || Date.now() < lockReleaseAt,
+    enabled
+  }
+}
+
+export function listSubscriptions(): AutoDownloadSubscription[] {
+  if (!deps) return []
+  const map = deps.store.get('autoDownloadSubscriptions') as Record<string, AutoDownloadSubscription>
+  return Object.values(map)
+}
+
+export function getSubscription(animeId: number): AutoDownloadSubscription | null {
+  if (!deps) return null
+  const map = deps.store.get('autoDownloadSubscriptions') as Record<string, AutoDownloadSubscription>
+  return map[String(animeId)] ?? null
+}
+
+export function setSubscription(
+  animeId: number,
+  enabled: boolean,
+  meta?: { malId: number; animeName: string }
+): AutoDownloadSubscription | null {
+  if (!deps) return null
+  const map = { ...(deps.store.get('autoDownloadSubscriptions') as Record<string, AutoDownloadSubscription>) }
+  const key = String(animeId)
+  if (!enabled) {
+    if (key in map) {
+      delete map[key]
+      deps.store.set('autoDownloadSubscriptions', map)
+    }
+    return null
+  }
+  if (!meta) {
+    const existing = map[key]
+    if (!existing) return null
+    return existing
+  }
+  // Stamp lastEnqueuedEpisodeInt to the current episodes_aired so we never backfill.
+  const cache = deps.store.get('shikimoriAnimeDetails') as Record<string, ShikiAnimeDetailsCacheEntry>
+  const cacheEntry = cache[String(meta.malId)]
+  const aired = cacheEntry?.details?.episodes_aired ?? 0
+  const sub: AutoDownloadSubscription = {
+    animeId,
+    malId: meta.malId,
+    animeName: meta.animeName,
+    subscribedAt: Date.now(),
+    lastEnqueuedEpisodeInt: aired,
+    lastCheckedAt: 0
+  }
+  map[key] = sub
+  deps.store.set('autoDownloadSubscriptions', map)
+  return sub
+}
+
+function pickTranslation(
+  episode: EpisodeDetail,
+  preferred: { type: string; author?: string } | null,
+  globalType: string
+): Translation | null {
+  const active = episode.translations.filter((t) => t.isActive === 1)
+  if (active.length === 0) return null
+
+  const byHeightDesc = (a: Translation, b: Translation): number => b.height - a.height
+
+  if (preferred) {
+    if (preferred.author) {
+      const exact = active
+        .filter((t) => t.type === preferred.type && t.authorsSummary === preferred.author)
+        .sort(byHeightDesc)
+      if (exact[0]) return exact[0]
+    }
+    const sameType = active.filter((t) => t.type === preferred.type).sort(byHeightDesc)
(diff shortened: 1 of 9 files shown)


src/main/auto-downloader.ts
```diff
@@ -0,0 +1,413 @@
...
+    if (mapDirty) {
+      deps.store.set('autoDownloadSubscriptions', map)
+    }
+
+    lastTickResult = result
+    deps.broadcast('auto-dl:tick-result', result)
+    return result
+  } finally {
+    tickRunning = false
+    lockReleaseAt = 0
```
mikkerlo: nit: resetting `lockReleaseAt = 0` in the finally block negates the 60s cooldown for manual triggers. If a cooldown period was intended, this reset should be removed.

src/main/auto-downloader.ts
```diff
@@ -0,0 +1,413 @@
...
+  if (!enabled) {
+    if (key in map) {
+      delete map[key]
+      deps.store.set('autoDownloadSubscriptions', map)
+    }
+    return null
+  }
+  if (!meta) {
+    const existing = map[key]
+    if (!existing) return null
```
mikkerlo, line 122: nit: if the Shikimori cache is empty at the moment of subscription, `aired` will be 0, potentially causing a full backfill on the next tick. Might be worth a small safety check or an inline refresh here if `aired === 0`.

mikkerlo (approved):
Technically solid and well-integrated auto-download feature. The forward-only logic and translation resolution are well-implemented. Architecture follows project standards.

LGTM after manual testing.

akvasha:
Addressed both review nits in a follow-up commit (43448357):

1. **`lockReleaseAt = 0` reset (line 411)** — Fixed. Dropped the reset from the `finally` block so the `REENTRANCY_LOCK_MS` deadline survives tick exit; manual re-triggers within 60s now correctly hit the locked branch instead of running back-to-back. `tickRunning` still flips back to `false` so the next tick can run after the cooldown elapses.

2. **Empty Shikimori cache on subscribe (line 122)** — Fixed. `setSubscription` is now async: when the cache is missing or older than `DETAILS_FRESHNESS_MS` (6h) at subscribe time, it awaits `refreshShikimoriDetails(malId)` before reading `episodes_aired`. If `aired` is still 0 after the refresh, the subscription is refused (returns `null`) and `AnimeDetailView` shows a small alert so the user retries rather than silently arming a full backfill. The IPC handler at `src/main/index.ts` was made async to match.

Typecheck is clean.

mikkerlo (approved):
Technically solid and well-integrated auto-download feature. The forward-only logic and translation resolution are well-implemented. Architecture follows project standards. LGTM after manual testing.

PR #80, opened 2026-05-07
feat: cleanup prompt when a show is marked Completed on Shikimori

### LINKED ISSUE #77: Prompt to clean up local files when a show is marked Completed on Shikimori

## Motivation

When a user marks a show "Completed" on Shikimori, the local files (often several GB per show) just sit on disk forever. The Hot/Cold storage feature helps for users who set it up, but it's manual and tied to *all* finished episodes, not the show-level "I'm done with this" signal.

A "you finished this show, want to clean it up?" prompt at the moment the rate flips to `completed` is the highest-signal moment to recover disk space — the user has explicitly said they're done. Today we ignore that signal entirely.

## Approach & Architecture

A new prompt-only cleanup flow, gated on the existing `shikimori:rate-updated` IPC broadcast. **No automatic action ever** — the user always confirms.

**Trigger paths:**
1. **Reactive** — when `shikimori:rate-updated` fires with `status === 'completed'` AND the previous status was *not* `completed` (avoid re-prompting on re-saves of an already-completed rate), AND the anime has any local files (`downloadedAnime[animeId]` exists). The renderer (top-level `App.vue`) listens, checks the gate, and shows a non-blocking toast: *"You finished <name>. Free up <X> GB? [Cleanup] [Keep] [Don't ask for this show]"*.
2. **Manual** — a "Cleanup files…" button in the existing AnimeDetailView header (next to Star / Auto-download from #76). Always available when the show has any local files; opens the same confirmation modal as the prompt.

**Cleanup action** (when confirmed):
- Reuses the existing `downloaded-anime-delete` IPC handler — it already removes the show from `downloadedAnime`, deletes the folder, and (in advanced storage mode) sweeps both hot and cold dirs.
- Also clears `downloadedEpisodes` entries for that `animeId` (already handled by the downstream cleanup paths but verify).
- Keeps `library` (starred) entry, `watchProgress`, `animePrefs`, `skipDetections`, and the Shikimori rate untouched — only files + the `downloadedAnime` marker go.

**Snooze map** (`autoCleanupSnoozedAnimeIds`, new electron-store key, `Record<animeId, true>`):
- "Don't ask for this show" persists in this map.
- Keyed by `animeId` so a future re-download of the same show would still respect it (intentional — user explicitly silenced).
- Reset via a "Reset cleanup prompts" button in Settings > Storage (lists snoozed shows, lets user un-snooze individually or all at once).

**Disk usage display**: the toast and modal both compute the show's on-disk footprint via a lightweight `cleanup:get-size` IPC (recursive `fs.stat` over the show folder, both hot and cold dirs). Formatted with the existing `formatBytes` utility.

**No completeness gate** — per scoping decision, the trigger fires regardless of how many episodes are downloaded vs. aired vs. watched. The "Completed" rate is the user's explicit signal; we trust it.

## UI / UX Considerations

**Toast (reactive trigger):**
```
┌──────────────────────────────────────────────────┐
│ ✓ You finished «Frieren: Beyond Journey's End» │
│ 14.2 GB on disk                                  │
│                                                  │
│ [Cleanup files] [Keep]   [Don't ask for this]    │
└──────────────────────────────────────────────────┘
```
Auto-dismisses after 30s if ignored (treated as "Keep" — no snooze, will re-prompt only if the rate is bumped to non-completed and back to completed, which is rare).

**Modal (Cleanup files… button or "Cleanup files" tap from toast):**
- Title: "Clean up <show name>?"
- Body: lists what will be deleted — *N video files, M subtitle files, <size> total, in <hot dir> and <cold dir>*. Plus a sub-line "Watch progress and your Shikimori rate will be kept."
- Buttons: "Delete files" (destructive style) / "Cancel".
- No checkbox for "also clear watch progress" in v1 — out of scope.

**AnimeDetailView header**: button appears next to existing actions only when `downloadedAnime[animeId]` exists. Disabled with a tooltip ("No local files for this anime") otherwise.

**Settings > Storage**: new "Cleanup prompts" subsection with the snooze list manager.

## Implementation Plan

   - [ ] New IPC `cleanup:get-size` (animeId, animeName) — returns `{ bytes, files }` summed across hot + cold dirs (using existing folder-resolution helpers from `downloaded-anime-delete`).
   - [ ] New IPC `cleanup:execute` (animeId, animeName) — delegates to existing `downloaded-anime-delete` logic + clears `downloadedEpisodes` entries.
   - [ ] New IPC `cleanup:get-snoozed` / `cleanup:set-snoozed` (animeId, snoozed: boolean) for the Settings list.
   - [ ] In the existing Shikimori rate-update path (around `src/main/index.ts:2942`/`:2977`), detect `status: 'completed'` transitions: compare against the rate snapshot held by the broadcast (the cached entry has the prior state) and broadcast a new `cleanup:prompt` event when the gate passes (not-already-completed → completed AND `downloadedAnime[animeId]` AND not in snooze map).
   - [ ] Defaults: `autoCleanupSnoozedAnimeIds: {}` in store schema.
   - *Note: update both `src/preload/index.ts` and `src/preload/types.d.ts`.*
   - [ ] Expose `cleanupGetSize`, `cleanupExecute`, `cleanupGetSnoozed`, `cleanupSetSnoozed`, `onCleanupPrompt`.
   - [ ] Top-level toast handler in `App.vue` listening on `cleanup:prompt`. Renders the toast component with three actions wired to `cleanupExecute` / no-op / `cleanupSetSnoozed(true)`.
   - [ ] New small modal component `CleanupModal.vue` shared between the toast's "Cleanup files" path and the AnimeDetailView button.
   - [ ] `AnimeDetailView.vue`: new "Cleanup files…" header button gated on `library-is-downloaded`, opens the modal.
   - [ ] `SettingsView.vue` Storage tab: new "Cleanup prompts" subsection rendering the snooze list with un-snooze controls.

## Files to Touch

- **Modify:** `src/main/index.ts` (rate-update detection, new IPC handlers, store schema)
- **Modify:** `src/renderer/src/App.vue` (toast wiring)
- **New:** `src/renderer/src/components/CleanupModal.vue`
- **Documentation:** `DESIGN.md` (new IPC channels, new store key, cleanup flow section)

## Testing Strategy

- **Reactive prompt fires once**: download a show, mark it `completed` on Shikimori from the AnimeDetailView. Verify toast appears with correct size + filename count.
- **Re-saves don't re-prompt**: open the modal, cancel. Re-save the rate at `completed`. Verify no toast.
- **Status transitions through completed**: flip rate to `completed`, then to `dropped`, then back to `completed`. Verify the second `completed` fires a new prompt (because the prior state was `dropped`, not `completed`).
- **No local files = no prompt**: mark a never-downloaded show as `completed`. Verify silence.
- **Snooze persistence**: click "Don't ask for this show", re-trigger via re-flip. Verify silence. Restart app, re-flip. Verify still silent.
- **Snooze reset**: in Settings, un-snooze the show, re-flip. Verify prompt returns.
- **Manual button**: open AnimeDetailView for a downloaded show that's not completed, click "Cleanup files…". Verify modal opens and works.
- **Hot + cold cleanup**: in advanced storage mode with both dirs populated, verify the cleanup removes from both.
- **Size accuracy**: cross-check the displayed "X GB" against `du -sh` of the show folder.
- **Cross-platform**: Windows (locked file warning?), Mac, Linux.

## Risks & Edge Cases

- **Race with active downloads**: if the user marks a show `completed` while episodes are still downloading, the cleanup would either fail (file locked) or kill an in-flight download. Solution: the modal queries the live download queue and shows a warning row if any episode for this `animeId` is `queued`/`downloading`/`paused`. The "Delete files" button is replaced with "Cancel downloads + delete files" in that case (chains `download:cancel-by-episode` then `cleanup:execute`).
- **Player has the file open**: on Windows, deleting an open .mkv fails. The modal can detect this via the existing `player:opened` lock (when #63 lands) or just surface the OS error and ask the user to close the player. Until #63, surface an informative error toast on `EBUSY`.
- **Star + Cleanup interaction**: starred shows are *not* removed from `library` by the cleanup. The library tab will continue to show the show with no "downloaded" indicator — desired.
- **Auto-merge / chapter inject in flight**: same as the active-downloads case; the modal's "anything in flight?" check should include the merge queue.
- **Prior status detection**: the broadcast must include the prior status, or the renderer must keep its own snapshot of the cached rate. Easier to add `prevStatus` to the existing `shikimori:rate-updated` broadcast payload (small payload bump, no schema break for consumers that ignore it).
- **Disk-size scan latency**: large cold dirs over a network share could take seconds to walk. The toast renders with a "calculating…" placeholder and updates when `cleanup:get-size` resolves. Modal blocks on the size only when opened directly (not from the toast, where the size is already cached).
- **Snooze lookup cost**: `Record<animeId, true>` lookups are O(1); no concerns at any realistic library size.

## Out of Scope

- **Auto-act modes** (auto-move, auto-delete) — explicitly prompt-only per scoping. If users hit "Cleanup" frequently we can revisit a "Cleanup all completed" Settings action.
- **Auto-move-to-cold integration** — cleanup deletes; it doesn't archive. If users want "move to cold instead of delete" as a third button, file a follow-up.
- **Watched-only cleanup** ("delete only watched episodes") — different feature; #54 already covered storage dashboard ideas, this is intentionally simpler.
- **Score-threshold or rating-based triggers** — out per scoping; could be a follow-up.
- **Periodic startup sweep** — out per scoping; rely on the rate-update event + manual button. Users who want a one-shot sweep of "all completed shows that still have files" can use the manual button per-show, or we file a follow-up issue if demand emerges.
- **Cross-show batch cleanup UI** — out for v1; one show at a time keeps the UX simple.
- **Restoring deleted files** — irreversible; the modal language should make this clear.

akvasha:
## Summary

Surfaces a non-blocking toast at the highest-signal moment for recovering disk space — the user just told Shikimori they're done with a show. No automatic deletion: every cleanup goes through an explicit confirmation.

Fixes #77.

## Changes

**Reactive trigger.** `shikimori:update-rate` broadcasts a new `cleanup:prompt` event when the status flips to `completed` and the prior status was not `completed` (including first-time rate creation, which previously had no transition signal). Fires on the offline-fallback path too, so offline status flips still prompt. Gated on `downloadedAnime[animeId]` and the new `autoCleanupSnoozedAnimeIds` map.

**Manual trigger.** New "Cleanup files…" button in the AnimeDetailView header (visible while the show has local files), opens the same modal.

**Shared `CleanupModal.vue`** lists file count + on-disk size, warns about in-flight downloads (chains `download:cancel-by-episode` → `cleanup:execute`), and preserves `library` / `watchProgress` / Shikimori rate when deleting.

**Settings → Storage → Cleanup prompts** manages the per-show "Don't ask for this show" snooze list (un-snooze + reset all).

**New IPC:** `cleanup:get-size`, `cleanup:get-active-downloads`, `cleanup:execute`, `cleanup:get-snoozed`, `cleanup:set-snoozed`, `cleanup:prompt`.

**New store key:** `autoCleanupSnoozedAnimeIds: Record<animeId, true>`.

**`cleanup:execute`** unifies what `downloaded-anime-delete` already did *plus* the per-anime sweep of `downloadedEpisodes` entries that the legacy handler skipped. Caches (`animeCache`, `skipFingerprintCache`, `skipDetections`) and the in-memory `fileCheckCache` are pruned for the show as well.

DESIGN.md: new IPC rows, new store key, and a "Completion-triggered cleanup" subsection under File Management.

## Behavior

- Re-saves of an already-`completed` rate: no prompt (only true transitions trigger).
- Show with no local files: no prompt.
- "Don't ask for this show" persists across restarts and survives re-downloads (intentional).
- Toast auto-dismisses after 30 s if ignored (treated as "Keep").
- Active-download case: modal switches the destructive button to "Cancel downloads + delete files".

## Test plan

- [x] `npm run dev` boots
- [x] Flip a downloaded show to Completed → toast appears with size + file count
- [x] "Cleanup files…" header button on a downloaded show opens the modal and deletes
- [ ] "Don't ask for this show" silences future prompts; Settings → Storage → Cleanup prompts un-snoozes
- [ ] First-time rate creation at Completed (no prior cached entry) also fires the prompt
- [ ] No prompt on re-save of an already-Completed rate
- [ ] No prompt for a never-downloaded show
- [ ] With downloads in flight: modal shows "Cancel downloads + delete files" path
- [ ] Hot + cold storage: cleanup removes from both dirs

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -149,7 +149,8 @@ const store = new Store({
       totalChecked: 0,
       faststartCount: 0,
       nonFaststartSamples: [] as { animeId: number; animeName: string; episodeInt: string; episodeLabel: string; filePath: string; firstNonFtypBox: string; checkedAt: number }[]
-    } as Mp4StreamingStats
+    } as Mp4StreamingStats,
+    autoCleanupSnoozedAnimeIds: {} as Record<string, true>
   }
 })
 
@@ -620,6 +621,74 @@ function isStarredAnime(animeId: number): boolean {
 
 function isCachableAnime(animeId: number): boolean {
   return isDownloadedAnime(animeId) || isStarredAnime(animeId)
+}
+
+function getDisplayName(anime: AnimeSearchResult): string {
+  return anime.titles?.romaji || anime.titles?.ru || anime.title
+}
+
+function isCleanupSnoozed(animeId: number): boolean {
+  const snoozed = store.get('autoCleanupSnoozedAnimeIds') as Record<string, true>
+  return !!snoozed[String(animeId)]
+}
+
+async function resolveSmotretFromMalId(malId: number, fallback: unknown): Promise<AnimeSearchResult | null> {
+  if (fallback && typeof fallback === 'object' && typeof (fallback as { id?: unknown }).id === 'number') {
+    return fallback as AnimeSearchResult
+  }
+  const map = store.get('malIdMap') as Record<string, AnimeSearchResult>
+  const cached = map[String(malId)]
+  if (cached?.id) return cached
+  try {
+    const resolved = await lookupByMalIds([malId])
+    return resolved[malId] ?? null
+  } catch {
+    return null
+  }
+}
+
+async function maybeBroadcastCleanupPrompt(
+  smotretAnime: unknown,
+  malId: number,
+  status: shikimori.ShikiUserRateStatus,
+  prevStatus?: shikimori.ShikiUserRateStatus
+): Promise<void> {
+  if (status !== 'completed' || prevStatus === 'completed') return
+  const sa = await resolveSmotretFromMalId(malId, smotretAnime)
+  if (!sa || typeof sa.id !== 'number') return
+  if (!isDownloadedAnime(sa.id)) return
+  if (isCleanupSnoozed(sa.id)) return
+  broadcastToAll('cleanup:prompt', {
+    animeId: sa.id,
+    animeName: getDisplayName(sa),
+    malId
+  })
+}
+
+function sumShowFiles(animeName: string): { bytes: number; files: number } {
+  const animeDirName = sanitizeFilename(animeName)
+  let bytes = 0
+  let files = 0
+  for (const dir of storageDirsForScan()) {
+    const showDir = path.join(dir, animeDirName)
+    let entries: fs.Dirent[]
+    try {
+      entries = fs.readdirSync(showDir, { withFileTypes: true })
+    } catch {
+      continue
+    }
+    for (const entry of entries) {
+      if (!entry.isFile()) continue
+      const lower = entry.name.toLowerCase()
+      if (!(lower.endsWith('.mkv') || lower.endsWith('.mp4') || lower.endsWith('.ass') || lower.endsWith('.srt'))) continue
+      try {
+        const stat = fs.statSync(path.join(showDir, entry.name))
+        bytes += stat.size
+        files++
+      } catch { /* ignore */ }
+    }
+  }
+  return { bytes, files }
 }
 
 type FileCheckResult = Record<string, { type: 'mkv' | 'mp4'; filePath: string; translationId?: number; author?: string }[]>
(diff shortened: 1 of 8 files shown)


src/main/index.ts
```diff
@@ -571,6 +572,74 @@ function isCachableAnime(animeId: number): boolean {
...
+): Promise<void> {
+  if (status !== 'completed' || prevStatus === 'completed') return
+  const sa = await resolveSmotretFromMalId(malId, smotretAnime)
+  if (!sa || typeof sa.id !== 'number') return
+  if (!isDownloadedAnime(sa.id)) return
+  if (isCleanupSnoozed(sa.id)) return
+  broadcastToAll('cleanup:prompt', {
+    animeId: sa.id,
+    animeName: getDisplayName(sa),
+    malId
```
mikkerlo, line 664: nit: `sumShowFiles` uses sync `fs` methods. While perfectly fine for typical anime folder sizes, using `fs.promises` would prevent any potential main-thread blocking if a user has a massive number of files in a directory.

mikkerlo (approved):
Excellent implementation of completion-triggered cleanup prompts. The feature is well-integrated into the Shikimori sync flow and provides a clean UX for disk management. Architecture and IPC patterns follow project standards.

LGTM after manual testing.

akvasha:
Thanks for the review! Pushed a follow-up commit:

1. **`sumShowFiles` → async / `fs.promises`** — converted `readdirSync`/`statSync` to `await fs.promises.readdir` + `Promise.all` over `fs.promises.stat`. The single callsite (`cleanup:get-size`) is now `async` too; `ipcMain.handle` returns the Promise to the renderer unchanged.

mikkerlo (approved):
Excellent addition to the storage management flow. The implementation is thorough, handling both hot/cold storage and all relevant caches. The UX is polished and respects user intent via the snooze mechanism. LGTM after manual testing.

PR #81, opened 2026-05-08
feat: pre-fetch the next episode while watching (#78)

### LINKED ISSUE #78: Pre-fetch the next episode while watching

## Motivation

When binge-watching a show, a user typically opens the player on episode N, watches it, then opens episode N+1. If episode N+1 is local, that's instant. If it's not, they wait for either the smotret-anime stream URL to fetch (slow) or a fresh download to enqueue (slower). Either way, every transition introduces a few seconds of friction that adds up over a long session.

Since the player already knows which episode is playing and which translation is active, we have everything needed to predict and pre-fetch the next episode while the user is still in the middle of the current one.

## Approach & Architecture

A renderer-driven pre-fetch trigger living inside `PlayerView.vue`. No new background worker; the player is the natural owner of "what's the next episode" knowledge — it already maintains `allEpisodes`, `activeEpisodeIndex`, and the active `translationId`.

**Trigger logic** — configurable via a new `prefetchNextEpisode` setting (`'off' | 'open' | 'time-5min' | 'progress-50'`, default `'progress-50'`):
- `'off'`: no-op.
- `'open'`: fire once on `loadedmetadata` of the current episode.
- `'time-5min'`: fire once when cumulative real-playback time crosses 5 min (reuse the same `cumulativePlaybackTime` accumulator that powers the watched-flag detection at `PlayerView.vue` ~line 538). Survives pauses; doesn't re-fire on rewind.
- `'progress-50'`: fire once when `currentTime / duration >= 0.5`. Re-evaluated on `timeupdate`.

Each fire computes `nextIdx = activeEpisodeIndex + 1`, exits early if there's no next episode, otherwise dispatches a single pre-fetch attempt for that episode. The "fired this episode" flag is keyed by `${animeId}:${episodeInt}` so navigating to the next episode resets the trigger for the *new* current episode.

**Pre-fetch action**:
1. Resolve the target translation: read the *current* player session's active translationId from the existing `activeTranslation` ref. Find the next episode's translation matching `(translationType, author)` of the current one. If no exact match (e.g. that author skipped the episode), fall back to `animePrefs[animeId]` resolution — same priority chain as the manual download path.
2. Probe quality via `probe-embed-quality` (same as manual path), build a `DownloadRequest`, call `download:enqueue`. The existing pipeline handles everything from there — progress, retries, auto-merge, hot/cold storage.
3. Skip silently (no toast, no UI noise — the player view stays clean) when any of:
   - `downloadedEpisodes[animeId:nextEpisodeInt]` already has an entry, or the file lookup confirms it's on disk.
   - The next episode is already in the live download queue (any state: queued, downloading, paused, failed — failed ones are the user's problem to retry, not ours to compete with).
   - `autoDownloadSubscriptions[animeId]` is set (issue #76 will catch this episode on its own tick — avoid double-queueing and let the auto-downloader's rate-limit + ordering policy own it).
4. **No abort on player close** — once enqueued, the download lives its own life. Rationale: if the user comes back tomorrow, the file is ready; if they don't, it cost a few hundred MB. Symmetric with how we treat manual downloads.

**Streaming + local both supported**: the trigger doesn't care whether the *current* episode is local or streamed. The pre-fetched *next* episode is always a download (we're pre-fetching to disk, not pre-buffering a stream). Streaming users benefit because next time they hit "Next", they get a local file instead of another stream session.

**Toast / status surface**: a tiny inline indicator in the player UI — a thin "↓ Pre-fetching ep N+1" line in the existing settings/info overlay (or a single-line status badge near the episode chip), only visible while the pre-fetch is in flight. It taps the existing `download:progress` broadcast filtered by the matching `(animeId, episodeInt)`. Not a toast — toasts mid-playback are intrusive.

## UI / UX Considerations

**Player UI**: a non-blocking status row appears in the top-right info area while the pre-fetch is downloading:
```
↓ Ep 7 · 42% · 4.2 MB/s
```
Disappears when the download completes (or fails — failures are silent at the player level; they show up in DownloadsView like any other failed download).

**Settings > Player**: new "Pre-fetch next episode" dropdown:
- Off
- On player open
- After 5 minutes of playback
- At 50% playback (default)

Help text: "Starts downloading the next episode in the background while you're watching the current one. Uses your active translation and quality. Subscribed shows are skipped."

## Implementation Plan

   - [ ] No new IPC channels — pre-fetch goes through the existing `download:enqueue` handler.
   - [ ] New store key `prefetchNextEpisode` defaulting to `'progress-50'`. Add to schema.
   - [ ] No new channels.
   - [ ] `PlayerView.vue`: new `prefetchFiredKeys` Set ref to track per-episode firing. Watch `currentEpisodeInt` to clear the prior episode's flag if needed (actually the Set persists across episodes — different keys, different fires). Add a `tryPrefetch()` function that runs the resolve → guards → enqueue flow.
   - [ ] Wire `tryPrefetch()` into the trigger event matching the current setting:
     - `'open'` → call from `loadedmetadata` handler.
     - `'time-5min'` → call from the existing playback-time accumulator when the threshold is crossed.
     - `'progress-50'` → call from the `timeupdate` handler.
   - [ ] Read setting on player mount + on a `prefetch-setting-changed` window event (dispatched from Settings) to live-update without remount.
   - [ ] Dedupe / guard logic: read `downloadedEpisodesGet(animeId)`, `download:get-queue` (one call, filter by animeId), `autoDlGetSubscription(animeId)` (from #76). All run sequentially before enqueueing — fast because everything is already cached in memory in main.
   - [ ] Pre-fetch status indicator: small reactive piece in the player UI listening on the existing `download:progress` broadcast, filtered to `(animeName, episodeLabel)` matching the pre-fetched item.
   - [ ] `SettingsView.vue` Player tab: new "Pre-fetch next episode" dropdown, dispatches `prefetch-setting-changed` on save.

## Files to Touch

- **Modify:** `src/main/index.ts` (store schema entry only)
- **Documentation:** `DESIGN.md` (new setting, brief mention in Player section)

## Testing Strategy

- **`'open'` mode**: open ep 5. Verify ep 6 enqueues immediately.
- **`'time-5min'` mode**: open ep 5, scrub to 4:00, watch for 1 min. Verify ep 6 enqueues at the 5-min mark of *real playback*, not wall clock.
- **`'progress-50'` mode**: open ep 5 (24 min), scrub to 12:01. Verify ep 6 enqueues. Scrub back to 5:00 then forward to 12:01 again. Verify it does NOT re-enqueue.
- **Single fire per episode**: in `'progress-50'` mode, scrub past 50% several times. Verify exactly one enqueue.
- **Episode change**: in `'progress-50'` mode, watch ep 5 past 50% (ep 6 enqueues). Navigate to ep 6 and watch past 50%. Verify ep 7 enqueues.
- **Translation match**: play ep 5 with `voiceRu/AniDub`. Verify ep 6 is queued with `voiceRu/AniDub` if available, else falls back per `animePrefs`.
- **Translation missing for next ep**: play with a translation that has no ep 6. Verify the fallback chain picks something else, or skips with a debug log if no candidate exists.
- **Already-on-disk skip**: have ep 6 on disk; play ep 5 past trigger. Verify no enqueue.
- **Already-queued skip**: manually queue ep 6; play ep 5 past trigger. Verify no duplicate enqueue.
- **Subscribed skip**: subscribe to the show via #76; play ep 5 past trigger. Verify ep 7 (or whatever's next aired) is left to the auto-downloader.
- **Streaming → next ep enqueues**: stream ep 5 (no local file). Verify ep 6 starts downloading at the trigger.
- **Last episode**: play the latest aired episode. Verify nothing is enqueued (no nextIdx).
- **Setting live-update**: change setting from `'off'` to `'progress-50'` while player is open and past 50%. Verify trigger fires immediately on the next `timeupdate`.
- **Off mode**: setting=`'off'`. Verify no enqueue under any condition.

## Risks & Edge Cases

- **Pre-fetch races with manual download**: user clicks Download on ep 6 in AnimeDetailView at the same moment the pre-fetch fires. The existing `enqueue` dedupe (it removes stale duplicate entries before re-enqueuing — see `download-manager.ts:267`) handles this; verify the deduped item carries the right translation metadata.
- **Concurrent downloads contention**: `concurrentDownloads` defaults to 2. If the user's already running 2 downloads, the pre-fetch sits at `queued` until one finishes. That's fine — it's still cheaper than waiting until the user clicks Next.
- **Speed throttle interaction**: `downloadSpeedLimit` is shared across downloads. The pre-fetch will compete with whatever else is running, which may slow the user's deliberate downloads. No special handling — same trade-off as any background download.
- **5-minute trigger and pause/seek**: the cumulative playback accumulator (already present) clamps `< 2s` deltas to ignore seek jumps and pauses. So 5 min = 5 min of *actual* watch time, which is the intent.
- **Stream URL freshness**: pre-fetched episode goes through embed at enqueue time, so URL is fresh. If the queue defers it for hours, the existing 416-retry path handles staleness.
- **Browser-level pre-load competing with the player**: the in-app download is plain HTTP fetch in main, totally orthogonal to the renderer's `<video>` MSE pipe. No bandwidth-fighting beyond the OS-level network share.
- **Ep N+1 doesn't exist on smotret-anime yet**: `getEpisode(animeId)` returns no row → resolver returns null → silent skip. No retry; the auto-downloader (#76) is the right place for "wait for it to appear."
- **Player closes before download finishes**: per scoping decision, the download keeps running. If users complain about "I closed the player and it kept downloading," we add an opt-in "cancel pre-fetch on player close" toggle as a follow-up.
- **Disk full**: existing download-manager error surface handles this; failed pre-fetch shows up in DownloadsView. No new disk-space gate per scoping.
- **HEVC + transcode-fallback show**: pre-fetch downloads the source MKV; the transcode fallback only activates when *playing*. Pre-fetch isn't affected.

## Out of Scope

- **Disk-space gate before enqueueing** — explicitly skipped per scoping. Rely on existing failure handling.
- **Cancel pre-fetch when player closes** — explicitly skipped per scoping. The download is committed once it starts.
- **Pre-fetch N+2, N+3, ...** — only one episode ahead in v1. A "pre-fetch lookahead = 1/2/3" setting could be a follow-up if users want deeper buffers.
- **Pre-fetch the previous episode for back-navigation** — niche.
- **Pre-buffering a streaming session** (not a download — actually pre-warming the embed stream URL into a buffer) — different feature; could pair well with #63 (watching downloads as they grow).
- **Smart triggers based on user history** (e.g. "user always finishes 90% of episodes, pre-fetch on open") — out for v1; predictable explicit triggers are easier to reason about.
- **Surface pre-fetch failures as toasts in the player** — silent + visible in DownloadsView is enough for v1.
- **Throttling pre-fetches separately from manual downloads** — uses the existing `concurrentDownloads` and `downloadSpeedLimit` settings without special priority.

akvasha:
## Summary

Closes #78. While the user is watching episode N, `PlayerView` opportunistically enqueues a download for episode N+1 so the next transition is instant.

- New setting `prefetchNextEpisode` (Settings > Player): `off` / `open` / `time-5min` / `progress-50` (default).
- Translation matching prefers exact `(type, author)` of the current episode, falling back to best quality of the same type.
- Routes through the existing `download:enqueue` — no new IPC channels.

## Behavior

| Trigger | Fires from |
|---|---|
| `open` | `onDurationChange` (when the new episode's duration becomes known) |
| `time-5min` | `trackProgressDelta` once `cumulativePlayTime` ≥ 300s — counts real playback only, not wall clock |
| `progress-50` | `onTimeUpdate` once `currentTime / duration ≥ 0.5` |

Each episode fires at most once via a `${animeId}:${nextEpisodeInt}` key, so seeking back/forward across the threshold doesn't re-enqueue. Concurrent re-entrant attempts (very common with `timeupdate`) are de-duped by an in-progress key set, but a guard-induced bail does **not** permanently block the key — state changes (e.g. user unsubscribes mid-session) allow a future retry.

### Skip rules (silent)

- Next episode already on disk (`downloadedEpisodesGet`).
- Next episode is in the live queue with a non-terminal status.
- For auto-download-subscribed shows: the next episode is strictly newer than the subscription's frozen `initialEpisodesAired` snapshot. Older episodes (pre-subscription backlog, deleted-and-needs-redownload) still pre-fetch.

## UX

- 2.5s `Pre-fetching episode N…` toast on enqueue (centered, near the resume toast).
- Persistent `↓ Ep N · X% · Y MB/s` chip in the title bar while in-flight, polled every 1s via `download:get-queue` (deliberately not `onDownloadProgress`, because `offDownloadProgress` would tear down AnimeDetailView's listener).
- Once enqueued, the download is **not** cancelled on player close — symmetric with manual downloads.

## Stutter-on-seek fix

When a prefetch is downloading to the same disk that ffmpeg is reading the source MKV from, unbuffered seeks stuttered visibly until the download completed (concurrent disk writes vs. seek-respawn reads). The player now:

1. On `seeking`, checks whether the target is inside the SourceBuffer's `buffered` ranges.
2. If outside (= ffmpeg respawn imminent → disk-heavy) and a prefetch is in-flight, calls `download:pause` on it.
3. Resumes 1.5s after the last `seeked` event (debounced — bursts of seeks coalesce).
4. In-buffer seeks and non-MSE playback don't pause anything.
5. On player close mid-debounce, fires the resume explicitly so the queue keeps draining.

## Subscription model change

`AutoDownloadSubscription` now also stores `initialEpisodesAired` (set to `episodes_aired` at subscribe time, never modified). The pre-fetch boundary uses this frozen value rather than the rolling `lastEnqueuedEpisodeInt`, so the auto-downloader's progress doesn't accidentally widen the "skip" window. Existing subscriptions without the field fall back to `lastEnqueuedEpisodeInt`.

## Files

- `src/main/index.ts` — store default, `AutoDownloadSubscription` field.
- `src/main/auto-downloader.ts` — set `initialEpisodesAired` on subscribe.
- `src/preload/types.d.ts` — sync `EpisodeGroup` (was missing `animeId`/`episodeInt`/`translationType`/`author`) and `AutoDownloadSubscription`.
- `src/renderer/src/components/PlayerView.vue` — trigger, guards, polling, indicator, toast, seek-aware pause.
- `src/renderer/src/components/SettingsView.vue` — Player tab dropdown + `prefetch-setting-changed` event.
- `DESIGN.md` — setting entry + Player section.

## Test plan

- [x] `npm run typecheck`
- [x] `progress-50` default: open ep 4 of a downloaded show, scrub past 50% — ep 5 enqueues exactly once, toast appears, chip tracks progress.
- [x] Skip rules verified: stale-on-disk metadata GC'd via removing the file → next attempt enqueues correctly.
- [x] In-buffer seek doesn't pause prefetch (chip keeps ticking); out-of-buffer seek pauses, resumes 1.5s after stable.
- [x] Subscription guard uses `initialEpisodesAired` (legacy subs fall back to `lastEnqueuedEpisodeInt`).
- [ ] Last episode → no enqueue (no `nextIdx`).
- [ ] Off mode → no enqueue, no chip.
- [ ] `time-5min` mode end-to-end.
- [ ] `open` mode end-to-end.


src/renderer/src/components/PlayerView.vue
```diff
@@ -557,6 +629,159 @@ function resetEpisodeTracking(): void {
   episodeOpenedAt = Date.now()
 }
 
+async function tryPrefetch(): Promise<void> {
+  if (prefetchSetting.value === 'off') return
+  if (!canNext.value) return
```
mikkerlo, line 699: nit: `prefetchFiredKeys` and `prefetchInProgressKeys` are reset on every `PlayerView` mount since they are inside `<script setup>`. This is correct for the intended session-based behavior, just noting it for clarity.

src/renderer/src/components/PlayerView.vue
```diff
@@ -557,6 +629,159 @@ function resetEpisodeTracking(): void {
...
+    // deleted) are still pre-fetched.
+    const sub = await window.api.autoDlGetSubscription(props.animeId)
+    if (sub) {
+      const nextNum = parseInt(nextEp.episodeInt, 10)
+      const boundary = sub.initialEpisodesAired ?? sub.lastEnqueuedEpisodeInt
+      if (Number.isFinite(nextNum) && nextNum > boundary) {
+        console.log(`[prefetch] skip: ep ${nextEp.episodeInt} aired after subscription (boundary=${boundary})`)
+        return
+      }
+    }
```
mikkerlo, line 724: Clever use of `sourceBuffer.buffered` to detect if a seek will cause disk-heavy ffmpeg respawn. This is a very targeted fix for the hardware contention issue.

src/preload/types.d.ts
```diff
@@ -514,8 +515,12 @@ interface DownloadProgressItem {
 interface EpisodeGroup {
```
mikkerlo, line 517: Good catch adding these fields to `EpisodeGroup`. They were indeed missing and are now critical for the player to identify its specific in-flight pre-fetch.

mikkerlo (approved):
Solid implementation of the next-episode pre-fetch feature. The logic for triggers, guards (especially the subscription boundary), and the disk contention fix for seeks is well-thought-out. It aligns perfectly with the existing architecture and IPC patterns. LGTM after manual testing.

Issue #84, opened 2026-05-18
Epic: Refactor project structure — decompose monoliths, shared IPC contract, test foundation


akvasha:
**Priority:** High | **Effort:** Large (multi-PR epic — tracked here, executed in small slices)

> **This is an epic / tracking issue, not a single PR.** It proposes a full project-structure refactor and breaks it into independently-shippable phases. Each phase below should become its own issue + PR. Behavior must stay identical at every step (pure restructuring; no feature changes). Please review the *strategy* and *phasing* here, then we open the per-phase issues.

---

## Motivation

The project works well but has grown by accretion. The structural debt is now actively slowing feature work and review:

| File | Lines | Problem |
|------|------:|---------|
| `src/main/index.ts` | **4833** | App lifecycle + 117 IPC handlers + Shikimori sync + anime cache + ffmpeg mgmt + downloads + cold storage + skip analysis + MSE streaming + syncplay wiring + auto-download, all in one file. `registerIpcHandlers()` alone is a single ~2000-line function. |
| `src/renderer/src/components/PlayerView.vue` | **3622** | God component: player + MKV remux + MSE streaming + Anime4K WebGPU + JASSUB subtitles + syncplay + keyboard + skip markers. 141 functions in one `<script setup>`. |
| `src/renderer/src/components/AnimeDetailView.vue` | **3113** | Episode list + translations + download orchestration + chronology + cleanup. 98 functions. |
| `src/renderer/src/components/SettingsView.vue` | **2909** | 5 unrelated settings tabs (General/Connectors/Merging/Player/Debug) in one component. 90 functions. |
| `src/preload/index.ts` | 469 | One flat `api` object with **185 methods**, no namespacing. |
| `src/preload/types.d.ts` | 735 | All shared IPC types in a single 24KB file. |
| `DESIGN.md` | **85 KB** | Single doc covering every subsystem; unwieldy to read or keep current. |

Concrete pain points this causes:
- **Merge conflicts**: nearly every feature touches `index.ts` and the big `.vue` files, so concurrent PRs collide constantly.
- **No test coverage**: zero unit/integration tests. `@playwright/test` is installed but there is no config and no test directory. Pure, testable logic (skip detection, filename parsing, queue consolidation, continue-watching merge) is buried inside the monoliths and untestable in isolation.
[...]

PR #86, opened 2026-05-18
Phase 1a: split domain types into src/shared/types (#84)

akvasha:
## Summary

First slice of **Phase 1** of the structure-refactor epic (#84). **Pure type relocation — zero runtime code, zero behavior change, zero renderer churn.** Independently revertible.

`src/preload/types.d.ts` was an 878-line grab-bag of every shared type. This moves the 50 domain interfaces/aliases into per-domain files, leaving `types.d.ts` as just the IPC surface.

## Changes

- **New** `src/shared/types/{anime,download,player,storage,skip,shikimori,syncplay,app}.ts` — domain-grouped type declarations.
- **Slimmed** `src/preload/types.d.ts` to only `interface Api`, `interface Window`, and the libass module shim. The `Api` shape is **byte-identical** to before.
- Types remain **ambient global declarations** (no `import`/`export`), so every existing call site keeps referencing them by bare name. Back-compat is automatic — no consumer changes.

The relocated types are global ambient (the file is a TS *script*, not a module). Both `tsconfig.node.json` and `tsconfig.web.json` already include `src/shared/**` (added in Phase 0), so main, preload, and renderer all resolve the same global types exactly as before. No `import` was added anywhere; the only renderer-visible contract (`window.api: Api`) is unchanged.

This deliberately does **not** yet introduce the typed channel contract or regenerate the preload (epic decisions 1–2) — those are the riskier slices 1b/1c, kept separate per the epic's small-slice mandate.

## Test plan

- [x] `npm run typecheck` (node + web projects) — green
- [x] `npm run build` — succeeds
- [x] `npm run test` — green

Part of #84 (Phase 1, slice 1a).


mikkerlo (approved):
This is a great, isolated slice for Phase 1. The relocation is completely transparent to the rest of the application since the types remain ambient globals.

Two minor observations:

1.  **File extensions (`.ts` vs `.d.ts`)**: Currently, these new files are created as `*.ts`. While it works because they only contain `interface` and `type` declarations, it's safer to name them `*.d.ts`. If someone accidentally adds a runtime statement (like `export const foo = 1` or an enum) to a `.ts` file, it ceases to be an ambient global declaration and becomes a module, which would instantly break the types across the app. Using `.d.ts` enforces that only type declarations can exist in these files.
2.  **Type Shadowing**: `src/main/smotret-api.ts` currently exports its own duplicate versions of `AnimeSearchResult` and `AnimeDetail`. Now that the domain types are cleanly separated into `src/shared/types/anime.ts`, you might consider cleaning up that shadowing in a later phase so `smotret-api.ts` just uses the global definitions.

LGTM. No blocking changes needed here.

akvasha:
Addressed your review (new commit on top, per our no-squash convention):

1. **`.ts` → `.d.ts`**: Done. All 8 relocated domain-type files (`anime`, `app`, `download`, `player`, `shikimori`, `skip`, `storage`, `syncplay`) renamed to `*.d.ts`, so the ambient-only contract is now structurally enforced — a stray runtime statement will no longer silently turn one into a module. Pure rename; type shapes unchanged, `typecheck` green. Also updated the now-stale `src/shared/types/*.ts` reference in the `src/preload/types.d.ts` header comment.
2. **`smotret-api.ts` shadowing**: Left as-is intentionally — you scoped this "to a later phase" and it's now tracked under the Phase 2 service-extraction work (#89), where `smotret-api.ts` will drop its duplicate `AnimeSearchResult`/`AnimeDetail` and use the globals.

PR #87, opened 2026-05-18
Phase 1b: shared IPC channel contract (#84)

akvasha:
## Summary

Second slice of **Phase 1** (#84). **Purely additive — nothing imports the new module yet, so behavior is byte-identical.** Introduces the single source of truth for IPC channel names (epic decision 1) and the disposable-subscription contract (epic decision 2).

> ⚠️ **Stacked on #86 (Phase 1a).** Base is `phase-1a-shared-types` so the diff is clean. I'll rebase onto `main` and retarget to `main` once #86 merges.

## Changes

- **New `src/shared/ipc/channels.ts`:**
  - `CHANNELS` (`as const`) — all **117** request/response channel names, domain-grouped, with `SCREAMING_SNAKE` keys deterministically derived from each literal.
  - `EVENT_CHANNELS` (`as const`) — all **27** broadcast channel names.
  - `RequestChannel` / `EventChannel` union types.
  - `Unsubscribe` + `EventSubscriber<T>` — the subscription contract slice 1c implements (returns a disposer that removes **only its own** listener; never `removeAllListeners`).
- **New `test/ipc-channels.test.ts`** — re-derives the channel set directly from `src/main/index.ts` + `src/preload/index.ts` and asserts the contract mirrors it **exactly** (drift guard), literals are unique, request/event are disjoint, and every key matches the derivation rule.

`channels.ts` is a new module imported by nothing in `main`/`preload`/`renderer`, so it's absent from the build graph and changes no runtime path. It only adds type-checked surface. Wiring `main` to register against `CHANNELS` and regenerating `preload` from it is the deliberately-separate, higher-risk **slice 1c**.

## Test plan

- [x] `npm run typecheck` — green
- [x] `npm run build` — succeeds (module not in graph)
- [x] `npm run test` — 7 tests green (incl. the exhaustive-mirror check: contract == channels actually used in main+preload)

Part of #84 (Phase 1, slice 1b).

### DIFF
--- a/src/shared/ipc/channels.ts
+++ b/src/shared/ipc/channels.ts
@@ -1,0 +1,250 @@
+// Single source of truth for IPC channel names (epic #84, decision 1).
+//
+// Channel names live here once, as an `as const` object, and are dereferenced
+// by symbol on both sides of the boundary — `main` registers handlers against
+// these constants and `preload` builds `window.api` from them (wired in the
+// Phase 1c slice). A rename becomes a compile error instead of a silent miss,
+// retiring the error-prone "update the same string literal in 4 files" rule.
+//
+// This module is intentionally additive: nothing imports it yet, so Phase 1b
+// is behavior-identical. The literals below are the exhaustive set extracted
+// from the existing `src/main/index.ts` + `src/preload/index.ts` (verified by
+// test/ipc-channels.test.ts).
+
+/**
+ * Request/response channels — renderer `ipcRenderer.invoke` (or fire-and-forget
+ * `send`) handled by a single `ipcMain.handle`/`ipcMain.on` in main.
+ */
+export const CHANNELS = {
+  // App
+  APP_VERSION: 'app:version',
+
+  // Updates
+  UPDATE_CHECK: 'update:check',
+  UPDATE_DOWNLOAD: 'update:download',
+  UPDATE_INSTALL: 'update:install',
+
+  // Auto-downloader
+  AUTO_DL_GET_ENABLED: 'auto-dl:get-enabled',
+  AUTO_DL_GET_STATUS: 'auto-dl:get-status',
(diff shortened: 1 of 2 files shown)


mikkerlo (approved):
This is an excellent implementation of the shared IPC contract. The introduction of `Unsubscribe` and `EventSubscriber<T>` is a crucial step towards fixing the fragile `removeAllListeners` pattern currently used in the renderer.

I have one specific observation regarding the drift-guard test (`test/ipc-channels.test.ts`):

1.  **Scanner Robustness (`broadcastToAll`)**: The regex `evtRe` scans for `webContents.send` and `ipcRenderer.*`. However, looking at `src/main/index.ts`, almost all broadcast events are sent using a custom `broadcastToAll('channel:name', ...)` helper (there are over 30 instances of this). Since `broadcastToAll` takes the literal string as its first argument, the current regex will miss any channel that is *only* broadcast via this helper and not explicitly listened to by `ipcRenderer.on` in `preload/index.ts`. You should update `evtRe` to include `broadcastToAll`:
    ```typescript
    const evtRe = /(?:webContents\.send|broadcastToAll|ipcRenderer\.(?:on|removeListener|removeAllListeners))\(\s*[`'"]([a-z][\w:-]+)[`'"]/g
    ```

Aside from that scanner tweak, the structure looks solid and ready for Phase 1c. LGTM.

akvasha:
Re: the `evtRe` scanner-robustness note —

That observation is fully addressed downstream in 1c (#88), which reworks this test: `evtRe` no longer exists; the single `IPC_CALL` scanner now covers all emit/listen forms **including `broadcastToAll`**. Acting on your point exposed a real gap — `broadcastToAll`'s 30 call sites in `main/index.ts` were *not* migrated (the 1c codemod regex never matched the helper), and 6 syncplay broadcast channels were missing from the contract entirely. All fixed in #88. No change needed on this PR.

PR #88, opened 2026-05-18
Phase 1c: wire main + preload to the IPC contract (#84)

akvasha:
## Summary

Final slice of **Phase 1** (#84). **Behavior byte-identical.** Every IPC channel string literal in `main` and `preload` now dereferences its `CHANNELS` / `EVENT_CHANNELS` constant from `@shared/ipc/channels` — the single source of truth introduced in 1b becomes load-bearing.

> ⚠️ **Stacked on #87 → #86.** Base is `phase-1b-ipc-contract`. I'll rebase down onto `main` and retarget as #86/#87 merge.

## Changes

- `src/main/index.ts` — **128** distinct channel literals → constants (`ipcMain.handle/on`, `webContents.send`, `sender.send`).
- `src/preload/index.ts` — **144** distinct channel literals → constants (`ipcRenderer.invoke/send/on/removeListener/removeAllListeners`).
- Added `import { CHANNELS, EVENT_CHANNELS } from '@shared/ipc/channels'` to both.
- `test/ipc-channels.test.ts` reworked for the post-migration invariant: **no raw IPC channel literal may remain** (single source of truth enforced) and **no contract constant is dead** (each referenced by symbol), keeping the uniqueness / disjoint / key-derivation checks.

The constant values are the exact original strings — guaranteed by the 1b test that every key is `value.replace(/[:-]/g,'_').toUpperCase()`. The codemod only substitutes the first argument of IPC calls; no handler logic, no `on*/off*` set, and **no `window.api` / `Api` shape** changes. The renderer is untouched.

## Explicitly deferred (not a regression)

The `on*/off*` → `subscribe`-returns-disposer rewrite and the `removeAllListeners` footgun fix are **out of scope here**: they change `window.api`'s shape and every renderer subscription call site, which Phase 1 forbids ("no renderer changes yet"). The epic assigns that to **Phase 4** (renderer state layer). 1b already ships the `Unsubscribe`/`EventSubscriber` contract those will implement.

## Test plan

- [x] `npm run typecheck` — green
- [x] `npm run build` — succeeds (`@shared` resolves in main+preload bundles)
- [x] `npm run test` — 8 tests green (incl. "no literals remain" + "no dead constants")
- [ ] Manual smoke (post-merge): app launches, a few IPC round-trips (search, library toggle, download enqueue, a broadcast like download progress) behave identically

Part of #84 (Phase 1, slice 1c — completes Phase 1's main/preload wiring).

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -1,4 +1,5 @@
 import { app, shell, BrowserWindow, ipcMain, dialog, Notification, protocol, net } from 'electron'
+import { CHANNELS, EVENT_CHANNELS } from '@shared/ipc/channels'
 import { join } from 'path'
 import { is } from '@electron-toolkit/utils'
 import Store from 'electron-store'
@@ -949,7 +950,7 @@ async function backgroundRescan(animeName: string): Promise<void> {
   if (newJson !== cachedJson) {
     fileCheckCache.set(animeName, result)
     for (const win of BrowserWindow.getAllWindows()) {
-      win.webContents.send('file:episodes-changed', animeName, result)
+      win.webContents.send(EVENT_CHANNELS.FILE_EPISODES_CHANGED, animeName, result)
     }
   }
 }
@@ -1083,7 +1084,7 @@ function ensureFfmpeg(win?: BrowserWindow): Promise<string> {
 
   const sendProgress = (status: string, progress?: number): void => {
     if (win && !win.isDestroyed()) {
-      win.webContents.send('ffmpeg:download-progress', { status, progress })
+      win.webContents.send(EVENT_CHANNELS.FFMPEG_DOWNLOAD_PROGRESS, { status, progress })
     }
   }
 
@@ -2304,14 +2305,14 @@ function dropSkipDetectionsForEpisode(animeId: number, episodeInt: string): bool
 }
 
 function registerIpcHandlers(): void {
-  ipcMain.handle('app:version', () => app.getVersion())
+  ipcMain.handle(CHANNELS.APP_VERSION, () => app.getVersion())
 
-  ipcMain.handle('update:check', async () => {
+  ipcMain.handle(CHANNELS.UPDATE_CHECK, async () => {
     try {
       const result = await autoUpdater.checkForUpdates()
       if (!result) {
         for (const win of BrowserWindow.getAllWindows()) {
-          win.webContents.send('update:status', {
+          win.webContents.send(EVENT_CHANNELS.UPDATE_STATUS, {
             status: 'error',
             error: 'Update check not available in development mode'
           })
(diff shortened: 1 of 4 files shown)


mikkerlo (approved):
This final slice perfectly wires up the single source of truth for the IPC channels. 

Regarding the test scanner (`test/ipc-channels.test.ts`), I noticed that `broadcastToAll` wasn't explicitly added to `IPC_CALL`. While the refactor cleanly converted `win.webContents.send` and `sender.send` into constants, the `broadcastToAll` helper inside `main/index.ts` still takes a channel string and wraps `webContents.send`. 

However, you updated the 30+ call sites of `broadcastToAll` to use `EVENT_CHANNELS.*` (e.g. `broadcastToAll(EVENT_CHANNELS.FILE_EPISODES_CHANGED, ...)`), so those are successfully decoupled from raw string literals. The fact that the test still passes means that `IPC_CALL` is correctly identifying the `webContents.send` *inside* the `broadcastToAll` definition itself (since it uses `channel: string`), but it means the test isn't technically validating the *call sites* of `broadcastToAll`. If someone were to add `broadcastToAll('raw-literal-here')` in the future, the test might miss it because `IPC_CALL` only looks for `.send(` and `.handle(` etc. 

It would be slightly safer to add `|broadcastToAll` to the `IPC_CALL` regex in the test to ensure all *usages* are caught, but it's a minor nit since the current codebase is fully migrated.

The deferred `on*/off*` rewrite makes sense for Phase 4. LGTM.

akvasha:
Addressed your review (new commit on top, per our no-squash convention).

Your suggestion to add `broadcastToAll` to the guard was framed as a minor nit "since the codebase is fully migrated" — it turned out the codebase was **not** fully migrated, so this is now a substantive fix:

1. **Guard hardening**: Added `broadcastToAll` to `IPC_CALL` in `test/ipc-channels.test.ts`. This immediately went red — correctly.
2. **30 un-migrated `broadcastToAll()` call sites**: The 1c codemod's regex never matched the `broadcastToAll` helper, so all 30 call sites in `src/main/index.ts` still used raw string literals, bypassing the single source of truth. All migrated to `EVENT_CHANNELS.*`.
3. **6 missing contract channels**: `syncplay:{connection-status,remote-state,room-users,room-event,remote-episode-change,trace}` were absent from `EVENT_CHANNELS` entirely — the 1b extraction regex never saw them because preload subscribes via the custom `syncplayOn()` wrapper, not `ipcRenderer.on`. Added them to `channels.ts` and migrated the 12 `syncplayOn/Off()` literals in `src/preload/index.ts`.

All substitutions are byte-identical channel strings and the new contract entries are additive, so runtime behavior is unchanged. `typecheck`, `lint`, `format:check`, and `test` (8/8) all green. This completes Phase 1c's single-source-of-truth mandate. The `on*/off*` → subscribe-disposer rework remains deferred to Phase 4 as planned.

mikkerlo (approved):
Excellent work. You caught exactly what I meant by the `broadcastToAll` scanner fix, and adapting the regex surfaced both the un-migrated broadcast calls and the hidden `syncplay` channels. This ensures the single source of truth is absolute.

All my previous concerns are fully resolved. LGTM!

PR #90, opened 2026-05-19
Phase 1b: shared IPC channel contract (#84)

akvasha:
> **Replaces #87**, which GitHub auto-closed when the Phase 1a base branch was deleted on merge of #86 (and refused reopen after the post-merge rebase). Same branch, same commit, now correctly based on `main`. @mikkerlo previously approved #87 — re-approval needed here.

## Summary

Phase 1b of the #84 refactor epic. Introduces a single source of truth for IPC channel names so main, preload, and (later) renderer all reference one typed registry instead of duplicated string literals. Purely additive — no call site is rewired yet (that is Phase 1c, #88).

## Changes

- **New `src/shared/ipc/channels.ts`**:
  - `CHANNELS` — request/response channels (`ipcMain.handle` ⇆ `ipcRenderer.invoke`), domain-grouped, `as const`.
  - `EVENT_CHANNELS` — broadcast/push channels (`webContents.send`/`broadcastToAll` → `ipcRenderer.on`).
  - `RequestChannel` / `EventChannel` union types derived from the registries.
  - `Unsubscribe` / `EventSubscriber<T>` — the disposable-subscription contract Phase 4 will use to retire the fragile `removeAllListeners` pattern.
  - Keys are mechanically derived: `channel.replace(/[:-]/g,'_').toUpperCase()`.
- **New `test/ipc-channels.test.ts`** — drift guard: asserts no duplicate literals, request/event sets disjoint, and the SCREAMING_SNAKE key-derivation invariant. (Significantly hardened in #88 to also prove every literal in main/preload is migrated.)

## Behavior

None. No runtime code imports the registry yet; this commit only adds the contract module and its tests. Behavior is byte-identical.

## Test plan

- `npm run typecheck` — green
- `npm run test` — green (drift-guard invariants hold)
- `npm run lint` / `npm run format:check` — clean

### DIFF
--- a/src/shared/ipc/channels.ts
+++ b/src/shared/ipc/channels.ts
@@ -1,0 +1,250 @@
+// Single source of truth for IPC channel names (epic #84, decision 1).
+//
+// Channel names live here once, as an `as const` object, and are dereferenced
+// by symbol on both sides of the boundary — `main` registers handlers against
+// these constants and `preload` builds `window.api` from them (wired in the
+// Phase 1c slice). A rename becomes a compile error instead of a silent miss,
+// retiring the error-prone "update the same string literal in 4 files" rule.
+//
+// This module is intentionally additive: nothing imports it yet, so Phase 1b
+// is behavior-identical. The literals below are the exhaustive set extracted
+// from the existing `src/main/index.ts` + `src/preload/index.ts` (verified by
+// test/ipc-channels.test.ts).
+
+/**
+ * Request/response channels — renderer `ipcRenderer.invoke` (or fire-and-forget
+ * `send`) handled by a single `ipcMain.handle`/`ipcMain.on` in main.
+ */
+export const CHANNELS = {
+  // App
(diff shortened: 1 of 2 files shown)


mikkerlo (approved):
Since this is the same commit as PR #87 rebased, my previous approval stands.

Just a reminder from my #87 review regarding the drift-guard test (`test/ipc-channels.test.ts`):
The regex `evtRe` currently scans for `webContents.send` and `ipcRenderer.*`. However, in `src/main/index.ts`, almost all broadcast events are sent using the custom `broadcastToAll('channel:name', ...)` helper. Because the regex misses this, it won't catch channels that are *only* broadcast via this helper if they lack an explicit listener. You should update `evtRe` to include `broadcastToAll`:
```typescript
const evtRe = /(?:webContents\.send|broadcastToAll|ipcRenderer\.(?:on|removeListener|removeAllListeners))\(\s*[`'"]([a-z][\w:-]+)[`'"]/g
```

LGTM.

PR #91, opened 2026-05-19
refactor: Phase 2 PR1 — extract pure leaf helpers from main/index.ts (#89)

akvasha:
## Summary

First PR of the Phase 2 structure-refactor (#89). Strangler-fig extraction of pure, dependency-free **leaf** helpers out of the 5.6k-line `src/main/index.ts` into focused modules, re-imported at the original call sites. **Behavior byte-identical** — no IPC, settings, UI, or logic changes.

Covers issue slices **2a** + **2b** (batched per the agreed 5-PR plan for #89).

## Changes

**Slice 2a — `src/main/lib/`**
- `errors.ts` — `errorCode` (walks the `cause` chain) + `isNetworkError` + `NETWORK_ERROR_CODES`
- `shikimori-queue.ts` — `QueuedShikimoriUpdate` / `ConsolidatedWorkItem` types + `consolidateQueue` (per-`malId` offline-queue collapse)
- `filename.ts` — `parseEpisodeFromFilename` + `FILENAME_EP_RE`

**Slice 2b — `src/main/streaming/`**
- `codec-strings.ts` — `avcCodecString`, `hevcCodecString`, `aacCodecString` + the private `reverseBits32` helper (RFC 6381 / MSE-compatible codec strings)

**Supporting**
- `src/main/index.ts` shrunk: moved blocks deleted, helpers re-imported. Unused-after-extraction imports (`errorCode`, `ConsolidatedWorkItem`) dropped.
- `DESIGN.md` File Map updated for the four new modules.
- New Vitest unit tests (`test/lib/**`, `test/streaming/**`) — 25 new assertions pinning behavior before later phases move call sites.

## Behavior

None changed. Pure code moved verbatim; call sites unchanged. The moved functions have no `electron`/store dependency (`errors.ts` only type-imports through `shikimori.ts`), so they unit-test without the electron mock.

## Scope / deferral

The "continue-watching merge" also listed under slice 2a is **deferred to Phase 3**: `HOME_GET_CONTINUE_WATCHING` is a store/network-coupled IPC handler with no cleanly isolatable pure leaf — a behavior-identical extraction is not possible without moving call sites (Phase 3 work). Confirmed with maintainer.

## Test plan

- `npm run typecheck` — green
- `npm run lint` — 0 errors (only pre-existing warnings)
- `npm run build` — green
[...]

Issue #89, opened 2026-05-19
Phase 2: extract leaf services from main/index.ts (#84)


akvasha:
**Priority:** High | **Effort:** Large (multi-PR — tracked here, executed in slices 2a–2g)

> **Phase 2 of the structure-refactor epic #84.** Tracking issue, not a single PR. Each slice below is its own small, behavior-identical PR. **Blocked on Phase 1 landing** (1a #86 → 1b #87 → 1c #88): every slice edits `src/main/index.ts`, which 1c rewrites — starting before Phase 1 merges would create a deep, conflict-prone stack.

## Motivation

`src/main/index.ts` is **5616 lines**: app lifecycle + `registerIpcHandlers()` (a single ~2400-line function, lines ~2307–4741) + Shikimori sync + anime cache + ffmpeg mgmt + cold storage + skip analysis + MSE streaming, all in one module. This blocks unit testing (pure logic is buried and reaches `electron-store` directly), causes constant merge conflicts, and makes review hard. Phase 2 extracts the **leaf** services/helpers behind their current interfaces — no IPC changes, no behavior changes — so later phases (routers, DI, decomposition, tests) have clean seams.

## Approach & Architecture

Strangler-fig, behavior byte-identical. Each slice moves code into a new module and re-imports it from `index.ts` (call sites unchanged). Pure helpers get unit tests as they land (pins behavior before Phase 3+ moves call sites). `StorageService` becomes an injected interface (epic decision 7) so downstream services never `import` the `electron-store` singleton — the Phase 0 in-memory fake already satisfies the contract. Slices are ordered by dependency and ascending risk; `store/` is foundational and precedes the services that depend on it.

| Slice | Extract | Source (≈ lines) | Risk | Dep |
|---|---|---|---|---|
| **2a** `src/main/lib/` pure helpers + unit tests | `errorCode`, `isNetworkError`, `consolidateQueue`, `parseEpisodeFromFilename`, continue-watching merge | 230, 245, 396, 1574, ~3970 | Low | — |
| **2b** `src/main/streaming/codec-strings.ts` + unit tests | `avcCodecString`, `hevcCodecString`, `aacCodecString` (pure) | 5021–5290 | Low | — |
| **2c** `src/main/store/` `StorageService` | typed interface + electron-store impl + `migrateWatchProgressV2`; route ~187 `store.*` calls (37 keys) | 1931 + pervasive | Med | — |
[...]

PR #92, opened 2026-05-19
refactor: Phase 2 PR2 — StorageService injected persistence boundary (#89)

akvasha:
## Summary

Second PR of the Phase 2 structure-refactor (#89), implementing slice **2c**. Decouples the main process and its leaf services (`shikimori`, `auto-downloader`) from the `electron-store` singleton by introducing an **injected `StorageService` interface**. **Behavior byte-identical** — no IPC, settings, UI, or logic changes.

Independent of PR #91 (slices 2a+2b). Both branch off `main`; the only overlap is the top-of-file imports in `src/main/index.ts`, a trivial merge.

## Changes

**New `src/main/store/` (4 files):**
- `types.ts` — `StorageService<S>` interface with schema-typed `get`/`set` overloads for literal keys (preserves the precise return types `electron-store` gave to all 187 call sites) and `string`-key overloads for the dynamic `get-setting` / `set-setting` IPC paths. Pure, electron-free; re-homed from the Phase 0 in-memory fake's note.
- `keys.ts` — `PERSISTED_STORE_KEYS` frozen 46-key tuple acting as a rename guard.
- `migrate.ts` — pure `migrateWatchProgressV2(store)` (one-shot `watchedAt` backfill, flag-guarded, idempotent). Unit-tested against the in-memory fake.
- `index.ts` — `createStorageService(defaults)` wraps `new Store<S>` and exposes `migrateWatchProgressV2()` as a method delegating to the pure function.

**`src/main/index.ts`:**
- Inline `defaults: {...}` lifted to top-level `STORE_DEFAULTS`.
- `const store = createStorageService(STORE_DEFAULTS)` replaces `new Store({...})`.
- **Compile-time rename guard:** an `_AssertExact` type-level check asserts `keyof typeof STORE_DEFAULTS & string` exactly equals `PersistedStoreKey`. Renaming a default key without updating the tuple breaks the build.
- Standalone `migrateWatchProgressV2()` deleted; call site uses `store.migrateWatchProgressV2()`. **All ~187 `store.*` call sites unchanged.**

**`src/main/shikimori.ts`, `src/main/auto-downloader.ts`:**
- `import type Store from 'electron-store'` → `import type { StorageService } from './store/types'`.
- `ensureFreshToken(store: Store)` → `ensureFreshToken(store: StorageService)`.
- `AutoDownloaderDeps.store` and two free-function store params: `Store` → `StorageService`.
- Both modules only use `get`/`set`, so the surface change is trivially compatible.

## Behavior

None changed. The `StorageService` schema-typed overloads return the exact same `S[K]` types `electron-store` did, so existing `as` casts and direct typed accesses (e.g. the `store.get('syncplay')` spread at line ~4638) all continue to compile and behave identically.

## Test plan

Full local quality gate (the gate that gated #91 in CI):
- `npm run typecheck` ✓
- `npm run lint` ✓ (0 errors, only pre-existing warnings)
- `npm run format:check` ✓
- `npm test` ✓ (13/13 — 5 new across `keys.test.ts` (2) and `migrate.test.ts` (3), plus existing in-memory + IPC tests)
- `npm run build` ✓

## Notes

- `electron-store` import is now confined to `src/main/store/index.ts`. Other main-process modules type-only-import `StorageService` from `./store/types` (pure, no electron). This is the prerequisite for slices 2d–2g (anime-cache, cold-storage, skip-analysis, streaming) which build on the injected boundary.
- Phase 0's `test/helpers/in-memory-storage.ts` already structurally satisfies `StorageService`. Merging its local interface declaration into a single `implements StorageService` is deferred (its existing 2-arg `get(key, default)` test predates this interface and is out of scope for 2c).
- Part of #89. No version bump (structure-only).

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -40,6 +40,10 @@ Renderer (Vue)  --ipcRenderer.invoke-->  Preload (bridge)  --ipcMain.handle-->  
 | `src/main/lib/shikimori-queue.ts` | `QueuedShikimoriUpdate`/`ConsolidatedWorkItem` types + pure `consolidateQueue` (per-malId offline-queue collapse) |
 | `src/main/lib/filename.ts` | Pure `parseEpisodeFromFilename` (episodeInt + ext from sanitized download filename) |
 | `src/main/streaming/codec-strings.ts` | Pure ffprobe→codec-string mappers: `avcCodecString`, `hevcCodecString`, `aacCodecString` (MSE-compatible RFC 6381 strings) |
+| `src/main/store/types.ts` | Injected `StorageService<S>` persistence interface (schema-typed get/set/has/delete); decouples services from `electron-store` |
+| `src/main/store/index.ts` | `createStorageService(defaults)` electron-store binding + `MainStorageService.migrateWatchProgressV2()` |
+| `src/main/store/keys.ts` | `PERSISTED_STORE_KEYS` frozen tuple — rename guard, compile-time-asserted against `keyof STORE_DEFAULTS` in `index.ts` |
+| `src/main/store/migrate.ts` | Pure `migrateWatchProgressV2(store)` (one-shot `watchedAt` backfill) |
 | `src/preload/index.ts` | contextBridge API exposure to renderer |
 | `src/preload/index.d.ts` | Shared TypeScript interfaces for IPC communication |
 | `src/renderer/src/main.ts` | Vue app entry point |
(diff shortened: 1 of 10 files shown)


mikkerlo (approved):
This is an exceptionally well-executed refactor slice. Introducing the `StorageService` interface perfectly sets up the codebase for unit testing without the `electron` module entanglement. 

I particularly like the compile-time guard (`_AssertExact`) between `STORE_DEFAULTS` and `PERSISTED_STORE_KEYS`. That's a great defensive mechanism against silent data-loss regressions.

Everything looks correct and strictly adheres to the byte-identical/no-behavior-change rule. LGTM.

PR #95, opened 2026-05-20
refactor: Phase 2 PR5 — extract ColdStorageService read-side surface (#89)

akvasha:
## Summary

Fifth PR of the Phase 2 structure-refactor (#89), implementing the **read-side surface of slice 2e** (cold-storage). Extracts the storage-mode path helpers and the `scanUsage` storage walk out of `src/main/index.ts` into a dedicated service. **Behavior byte-identical.**

**Stacked on #94 (PR4 — SkipAnalysisService).** Each predecessor PR in the chain must be retargeted to `main` *before* its branch is deleted (memory: stacked-PR base branches).

## What's extracted (`src/main/services/cold-storage/index.ts`)

- **Path helpers**: `getDownloadDir` / `getColdStorageDir` / `isAdvanced` / `dirsForScan` — pervasively used throughout `index.ts` (~41 call sites routed through `coldStorageService`).
- **`scanUsage`**: the `StorageUsage` walk — enumerates hot+cold roots, parses filenames, classifies bytes per anime/episode/bucket, emits `STORAGE_USAGE_PROGRESS` via injected broadcast.
- **Types**: `StorageUsage`, `StorageAnimeUsage`, `StorageEpisodeUsage`, `UsageEpisodeFiles` re-homed into the service module.

**Deps-injected:** `store`, `downloadsFallbackDir` (`app.getPath('downloads')`), `sanitizeFilename`, `parseEpisodeFromFilename`, `broadcast`, `usageProgressChannel`.

## What stays in `index.ts` (deferred)

- `deleteEpisodeFiles`, `episodeHasInProgressDownload`, `pruneDownloadedEpisode`, `episodeFileExists`
- `moveFileToCold`, `moveEpisodeToColdStorage`, `moveAllFilesToColdStorage`
- `CleanupCandidate`, `findCleanupCandidates`, `cleanupRunning`, `runWatchedCleanup`

Issue #89 explicitly calls out **manual spot-check** for the move/cleanup ops (they mutate user files on disk). Isolating them into their own diff gives that risk surface a focused review path and a clean bisect point if anything regresses. They'll follow as a stacked PR.

## Tests

`test/services/cold-storage.test.ts` — 7 path-helper cases against `InMemoryStorage`:
- Simple-mode fallback to `<downloads>/anime-dl` when no `downloadDir`
- `downloadDir` preferred over fallback in simple mode
- `hotStorageDir` preferred over `downloadDir` in advanced mode
[...]

PR #94, opened 2026-05-20
refactor: Phase 2 PR4 — extract SkipAnalysisService leaf surface (#89)

akvasha:
## Summary

Fourth PR of the Phase 2 structure-refactor (#89), implementing the **leaf surface of slice 2f** (skip-analysis). Extracts the non-stateful, deps-injectable parts of the skip-analysis subsystem out of `src/main/index.ts` into a dedicated service. **Behavior byte-identical.**

**Stacked on #93 (PR3 — AnimeCacheService).** Base is `phase2-pr3-anime-cache-service`. Each predecessor PR in the chain must be retargeted to `main` (or the next surviving base) *before* its branch is deleted (memory: stacked-PR base branches).

## What's extracted (`src/main/services/skip-analysis/index.ts`)

- `normalizeDetections` — pure: coerce legacy detection rows to a `local` algorithm block with defaults
- `pruneCacheForEpisode`, `pruneCacheForAnime` — store-only fingerprint-cache GC
- `sweepFingerprintCache` — reconcile cache vs. disk; drops orphans
- `dropDetectionsForAnime`, `dropDetectionsForEpisode` — store mutators + broadcast
- `buildAutoSkipEpisodes` — scan-based `EpisodeInput[]` construction (mkv preferred, sorted)

**Deps-injected:** `store`, `scanEpisodeFiles`, `sanitizeFilename`, `broadcast`, `signatureUpdatedChannel`. All fs / store side effects flow through deps.

## What stays in `index.ts` (deferred)

- `autoSkipDebounce` / `autoSkipQueue` / `autoSkipDraining` state
- `scheduleAutoSkipAnalysis` / `enqueueAutoSkipAnalysis` / `drainAutoSkipQueue`
- `runSkipAnalysisInternal` + `broadcastSkipProgress` + `currentSkipAnalysis` state
- `runStreamSkipDetectionInternal` + `currentStreamSkipDetections` state

These entangle the **auto-skip queue with stream-skip detection** through shared module-level state (`currentSkipAnalysis`). A clean extraction requires resolving that entanglement first, which is non-trivial and would extend the slice beyond a behavior-identical move. The PR4 surface is the largest behavior-identical chunk that lifts out without touching that state machine; the queue/drain extraction is queued as a follow-up.

## Tests

`test/services/skip-analysis.test.ts` — 13 cases against `InMemoryStorage`:
- `normalizeDetections` (3): null pass-through, local pass-through, defaults filled
[...]

PR #93, opened 2026-05-20
refactor: Phase 2 PR3 — extract AnimeCacheService (#89)

akvasha:
## Summary

Third PR of the Phase 2 structure-refactor (#89), implementing slice **2d**. Extracts the anime-cache / poster-cache logic out of `src/main/index.ts` into a dedicated service consumed through dependency injection. **Behavior byte-identical** — no IPC, settings, UI, or logic changes.

**Stacked on #92 (PR2 — StorageService).** Base is `phase2-pr2-storage-service`. When PR2 merges to `main`, retarget this PR's base to `main` *before* deleting PR2's branch (memory: stacked-PR base branches).

> ℹ️ Originally batched with slice 2f (skip-analysis) in the agreed 5-PR plan. Split for reviewability — skip-analysis becomes its own PR4 stacked on this one. Net result: 6 PRs instead of 5.

## Changes

**New `src/main/services/anime-cache/index.ts`**
- `AnimeCacheEntry` interface re-homed here from `index.ts`.
- `AnimeCacheServiceDeps`: `store` (`StorageService`), `userDataDir` (for `poster-cache/` subdir), `fetchPoster` (smotret-anime poster fetcher), `isCachable` (index-owned predicate: anime in library or downloaded).
- `AnimeCacheService` surface: `getEntry` / `setEntry` / `deleteEntry` / `ensureEntry` / `updateAnimeDetail` / `updateEpisode` / `updateQualityProbe` / `cachePosterImage` / `getPosterCacheDir` / `getPosterCachePath` / `cleanupStale`.
- `createAnimeCacheService(deps)` returns the service; pure-ish functions inside, all fs/store side effects flow through injected deps.

**`src/main/index.ts`**
- 11 standalone cache functions + the local `AnimeCacheEntry` interface deleted (~103 lines).
- Single `animeCacheService = createAnimeCacheService({...})` instantiated next to `smotretApi`.
- 26 call sites routed through the service (`getCacheEntry` → `animeCacheService.getEntry`, `updateAnimeDetailCache` → `animeCacheService.updateAnimeDetail`, `cleanupStaleCache` → `animeCacheService.cleanupStale`, etc.).
- Removed now-unused `EpisodeDetail` type import.

## Tests

`test/services/anime-cache.test.ts` — 8 cases against `InMemoryStorage` + real fs in an `os.tmpdir()` sandbox:
- `setEntry` / `getEntry` persistence
- `ensureEntry` returns default without writing
[...]

PR #96, opened 2026-05-20
ci: trigger check.yml on PRs to any base branch (#89 stack)

akvasha:
## Summary

Drop the \`pull_request.branches: [main]\` filter from \`.github/workflows/check.yml\` so the \`quality\` + \`build\` + \`comment-artifacts\` jobs fire on PRs targeting *any* base branch, not just \`main\`.

## Why

The Phase 2 refactor (#89) ships as a linear stack of 5 PRs (#91, #92, #93, #94, #95). PR3+ target \`phase2-pr*-*\` branches, so the existing filter prevented CI from running on them at all (literally \"no checks reported\" on the PR page). Each stacked PR still needs the full local quality gate — but losing the CI signal for stacked work is worth fixing once at the workflow level.

## Behavior change

- **Before:** CI only ran on PRs whose **base** is \`main\`.
- **After:** CI runs on every PR, regardless of base branch.

## Cost

Some duplicate work as a stacked PR rebases forward (the same commits get checked once per PR they appear in). \`quality\` is cheap (typecheck + lint + format + tests + smoke) and the build job already has a \`paths-filter\` gate that skips it when only docs change. Net: small.

## Test plan

The next push to PR #93 / #94 / #95 will trigger CI for the first time — that's the verification.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -2,8 +2,6 @@ name: Check Build
 
 on:
   pull_request:
-    branches:
-      - main
   issue_comment:
     types: [created]
 


mikkerlo (approved):
This is a necessary and pragmatic change to support the stacked PR workflow. Dropping the branch filter ensures we get the vital CI signals for each isolated slice before they merge into their respective base branches.

LGTM.

Issue #111, opened 2026-05-21
Phase 4: renderer state layer (Pinia) + disposable subscription contract (#84)


akvasha:
**Priority:** High | **Effort:** Large (multi-PR — tracked here, executed in slices 4a–4e)

> **Phase 4 of the structure-refactor epic #84.** Tracking issue, not a single PR. Each slice below is its own small, behavior-identical PR. Phase 3 (#102) has landed, so `src/main/index.ts` is now bootstrap-only and every `ipcMain.handle` lives in `src/main/ipc/*.ipc.ts`. Phase 4 finally crosses the contextBridge — it carries Phase 1's deferred subscription-contract migration (decision 2) into preload + renderer, and replaces `App.vue`'s prop-drilled cross-view state with Pinia stores.

## Motivation

After Phase 3 the *main process* has clean seams (App controller, per-domain routers, injected `StorageService`, factory-closure services). The renderer hasn't moved since the epic started, and two structural problems block Phase 5's God-component decomposition:

1. **`App.vue` is a 580-line bag of prop-drilled state.** It owns `animeByView` / `animeHistoryByView` / `focusEpisodeIntForAnime` (per-view selection stacks), `animePrefs` (per-anime translation+author selection), `shikimoriLoggedIn`, `playerState` (overlay state for `PlayerView`), `shortcuts`, `ffmpegDownloading`/`ffmpegProgress`, and the cleanup-toast/cleanup-modal flow. Every cross-view interaction (open anime, save prefs, open player, ack ffmpeg progress) flows through `App.vue` as props + emits because there is no state layer. Phase 5 cannot decompose `AnimeDetailView` (3651 lines), `PlayerView` (4265 lines), or `SettingsView` (3443 lines) into sub-components until their cross-cutting state has somewhere to live — otherwise the prop-drill just grows another level.
[...]

PR #112, opened 2026-05-22
refactor: Phase 4 slice 4a — preload subscribe-returns-disposer (#111)

akvasha:
## Summary

Apply the Phase 1 decision-2 subscription contract (`Unsubscribe`, `EventSubscriber<T>`) across the contextBridge into the renderer. Every `on*`/`off*` pair on `window.api` collapses into a single `on*` that returns an `Unsubscribe` disposer removing **only the caller's** listener — no more `removeAllListeners` clobbering other subscribers on the same channel. Slice 4a of Phase 4 (#111); part of refactor epic #84.

This is pure plumbing — channel names, payload shapes, and main-process behavior are byte-identical.

## Changes

**Preload (`src/preload/`):**
- New `src/preload/subscribe.ts` exports a generic `subscribe<T>(channel): EventSubscriber<T>` helper. Each call registers a dedicated `ipcRenderer.on` listener and returns a closure invoking `ipcRenderer.removeListener` against that exact listener.
- `src/preload/index.ts` rewrites all **33** broadcast subscriptions to `subscribe<T>(EVENT_CHANNELS.X)` (or, for the one multi-arg channel `FILE_EPISODES_CHANGED`, an inlined variant with the same disposer shape). Deletes the `syncplayWrappers` map and the `syncplayOn`/`syncplayOff` helpers — closure ownership of the wrapped listener obsoletes the external Map. Deletes all **28** `ipcRenderer.removeAllListeners(…)` calls and every `off*` key.
- `src/preload/types.d.ts` adds ambient `type Unsubscribe = () => void`, changes every `on*` return type from `void` to `Unsubscribe`, deletes all 33 `off*` declarations, and drops the now-unused callback-identity arg from the 6 syncplay `on*`s.

**Renderer (`src/renderer/`):**
Migrated **7 files** to capture each `on*` return as `let unsubX: Unsubscribe | null = null` and invoke it in `onBeforeUnmount`/`onUnmounted` (and inline branches for the self-unsubscribing handler in `SettingsView.testSyncplayConnection`):
- `App.vue` — 2 disposers (ffmpeg-download progress, cleanup prompt)
- `AnimeDetailView.vue` — 10 disposers (download progress, file-episodes, 4× shikimori, sync status, 2× skip-detector, chapter-inject)
- `DownloadsView.vue` — 1 disposer (download progress)
- `HomeView.vue` — 2 disposers (shikimori rate/refreshed)
- `PlayerView.vue` — 12 disposers (5 player-stream, 6 syncplay, 1 skip-detector); 6 stored handler refs replaced with disposers
- `SettingsView.vue` — 8 disposers + rewrite of `testSyncplayConnection`'s subscribe/inner-unsubscribe flow to a captured closure disposer
- `ShikimoriView.vue` — 2 disposers (shikimori rate/refreshed)

**Tests:**
- `test/preload/subscribe.test.ts` (new, 3 tests) asserts the contract by stubbing `ipcRenderer`: delivery works, the returned disposer removes only the caller's listener (subscribe twice, dispose one, the other still fires), and subscriptions stay isolated per channel.

**Docs:**
- `DESIGN.md` gains a "Broadcast subscription contract" section above the IPC handlers table, summarising the new pattern.

## What's unchanged

- All 117 request channels and 33 event channels — names and payloads frozen.
- Main process — no edits under `src/main/`.
- Behavior in the renderer — subscriptions still fire the same callbacks, just with disposable handles.

## Behavior

No user-visible change. The pattern enforces per-listener removal, so two views subscribing to the same broadcast no longer fight over a global slot — but the previous `removeAllListeners` behavior was only ever wrong as a latent risk, not a current bug, so no behavior actually shifts in shipped flows.

## Test plan

- [x] \`npm run typecheck\` — green
- [x] \`npm run lint\` — 0 errors, 8 pre-existing warnings unchanged
- [x] \`npm run format:check\` — clean
- [x] \`npm run test\` — 100/100 (incl. 3 new contract tests)
- [x] \`npm run build\` — green
- [x] \`grep -rn "window.api.off\\|removeAllListeners" src/preload/ src/renderer/\` — only the contract-doc comment in subscribe.ts matches
- [ ] Manual: SettingsView mount/unmount; DownloadsView progress; HomeView Shikimori rate edit; PlayerView local MKV open + close; SettingsView "Test connection" against unreachable host (handler self-unsubscribes via disposer, no leak)

## Out of scope (later 4 slices of #111)

- Pinia install / \`stores/\` directory — slice 4b
- \`App.vue\` state migration to stores — slices 4b–4d
- \`App.vue\` final shrink to routing+layout — slice 4e
- CI grep gate forbidding \`window.api.off\` / \`removeAllListeners\` — slice 4e

Part of #111. Epic: #84.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -327,6 +327,16 @@ LibraryView shows both with indicators:
 
 ## IPC Handlers
 
+### Broadcast subscription contract (Phase 4 slice 4a, #111)
+
+`send`-direction event channels expose `on*` subscribers on `window.api` of shape
+`EventSubscriber<T> = (listener) => Unsubscribe` (see `src/shared/ipc/channels.ts`).
+Each call registers a dedicated listener and returns a disposer that removes only
+that listener — never `ipcRenderer.removeAllListeners`, which would clobber every
+other subscriber on the channel. Renderer callers capture the returned `Unsubscribe`
+and invoke it in `onBeforeUnmount`/`onUnmounted` (or store disposal in Phase 4b+).
+There are no `off*` methods on `window.api`.
+
 | Channel | Direction | Purpose |
 |---------|-----------|---------|
 | `validate-token` | invoke | Validate API token against smotret-anime.ru |
--- a/src/preload/index.ts
+++ b/src/preload/index.ts
@@ -1,34 +1,6 @@
 import { contextBridge, ipcRenderer, type IpcRendererEvent } from 'electron'
-import { CHANNELS, EVENT_CHANNELS } from '@shared/ipc/channels'
-
-// Per-channel map of {original-callback → wrapped-handler} so the matching
-// off* helper can call ipcRenderer.removeListener with the actual wrapped
-// function (the one we registered on ipcRenderer.on). Without this map,
-// the only way to remove a listener is removeAllListeners — which would
-// also rip out listeners installed by other components in the same view.
-type SyncplayHandler<T> = (data: T) => void
-type SyncplayWrapped = (event: IpcRendererEvent, data: unknown) => void
-const syncplayWrappers = new Map<string, Map<SyncplayHandler<unknown>, SyncplayWrapped>>()
-
-function syncplayOn<T>(channel: string, cb: SyncplayHandler<T>): void {
-  const wrapped: SyncplayWrapped = (_event, data) => cb(data as T)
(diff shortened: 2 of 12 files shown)


mikkerlo (approved):
LGTM after manual testing

PR #125, opened 2026-05-23
refactor: Phase 5 slice 5c — EpisodeList + EpisodeRow components (#118)

akvasha:
## Summary

Fifth and final sub-slice of **Phase 5 slice 5b/5c** (#118 / epic #84). Stacked on top of #124. Extracts the central episode list rendering — the densest reactivity surface in the entire app — into two components.

`AnimeDetailView.vue`: **1525 → 890 lines** (~635-line shrink).

> **Base branch is `phase5-slice-5b4-shiki-skip-panels`** — diff against `main` retargets itself once #124 lands. Per [[feedback_stacked_pr_base_branch]] I'll switch the base to `main` *before* each upstream branch is deleted on merge.

### `src/renderer/src/components/detail/EpisodeList.vue` (138 L)

Pagination row + `v-for` of `EpisodeRow` + the "Loading episodes…" status.

- **Props**: `playerMode: 'system' | 'builtin'`, `translationType: string`
- **Injects**: `useEpisodeList` via `EpisodeListKey` for `filteredEpisodes`, `episodeRows`, `loadingEpisodes`, `currentPage`, `totalPages`, `isPaginated`, `goToPage`.

### `src/renderer/src/components/detail/EpisodeRow.vue` (537 L)

Single episode row — the most reactive markup in the app.

- **Props**: `row: EpisodeRow`, `playerMode`, `translationType`
- **Injects**: `useEpisodeList` + `useEpisodeDownloads` via their keys
- **Renders**:
  - Episode name + queued lock badge OR translation picker (`<select>` with optgroups sorted by active type first, then by quality desc; `⬇` prefix on downloaded options)
  - Download/merge status: `merging → merge-failed → downloading (progress bar + speed + ETA) → queued → paused → failed`
  - Watched/progress/type/quality/file badges
  - Action buttons: Open/Folder/Delete (when file exists), Cancel (active download or merging), Play (built-in player + no file), Download (no file + not locked).

### `src/renderer/src/components/detail/translation-types.ts` (22 L)

Shared `TRANSLATION_TYPES` array (subRu / subEn / voiceRu / voiceEn / raw, each with `label`/`short`/`color`) + `typeChip(type)` helper. Used by the parent controls bar AND EpisodeRow — avoids the previous duplication.

### `detail/keys.ts` extended

Two new typed `InjectionKey` symbols: `EpisodeListKey` and `EpisodeDownloadsKey`.

### `AnimeDetailView.vue` changes

[...]

Issue #118, opened 2026-05-22
Phase 5: renderer God-component decomposition (#84)


akvasha:
**Priority:** High | **Effort:** Large (multi-PR — tracked here, executed in slices 5a–5f)

> **Phase 5 of the structure-refactor epic #84.** Tracking issue, not a single PR. Each slice below is its own small, behavior-identical PR. Phase 4 (#111) has landed, so the renderer state layer (Pinia stores) and the disposable subscription contract are in place — `App.vue` is now ~145 lines and every cross-view broadcast is store-owned. Phase 5 finally cracks open the three remaining renderer monoliths.

## Motivation

After Phase 4 the cross-cutting renderer state has somewhere to live, but the three God components still hold every feature inside one `<script setup>` each:

| File | Lines | Top-level fns | Surface |
|---|---:|---:|---|
| `src/renderer/src/components/PlayerView.vue` | **4260** | 194 | MSE session + ffmpeg remux/transcode + Anime4K WebGPU + JASSUB subtitles + syncplay client + keyboard + skip markers + control bar + quality/translation menus |
| `src/renderer/src/components/AnimeDetailView.vue` | **3624** | 136 | Episode list + per-episode actions + translation picker + chronology panel + cleanup-modal wiring + download orchestration |
| `src/renderer/src/components/SettingsView.vue` | **3435** | 147 | 8 unrelated tabs in one file: general, storage, connectors, merging, player, shortcuts, watch-together, debug |

Concrete pain caused by leaving these intact:
- **Merge conflicts** on every feature PR that touches the player, episode list, or any setting tab — these three files are the next-highest collision surface after `main/index.ts` (now fixed in Phase 3).
- **Untestable logic.** Phase 4 unblocked Pinia store tests, but the heart of the app — the MSE/remux/transcode state machine — is still buried in a 4260-line `<script setup>` next to WebGPU shaders and DOM event handlers. Decision 5 of the epic (headless `useMsePlayer` extracted *first*, before any cosmetic decomposition) targets exactly this seam.
- **Settings tabs share no state** but compile/render as one component; switching a single tab repaints the whole script.
- **AnimeDetailView episode-list reactivity** is currently coupled to translation picker + download state in the same scope — Phase 7 cannot meaningfully test the episode list while it's a 3624-line bag.

[...]

PR #126, opened 2026-05-23
refactor: Phase 5 slice 5d.1 — useMsePlayer composable groundwork (#118)

### LINKED ISSUE #127: MKV streaming stutter + audio dropout on WSL (pre-existing, not on native Windows)
## Symptom

When playing a downloaded `.mkv` in the built-in player under **WSL** (WSLg/Linux build of Electron running on Windows), a quick back-to-back seek pattern triggers:

- Continuous micro-stutter on the video (repeated `video 'waiting' readyState=1` events even when the `SourceBuffer.buffered` range shows >40 s of lead ahead of the playhead).
- **Audio dropout** — sound disappears entirely while the (stuttering) video continues. Sometimes recovers, sometimes stays muted until a re-seek.

Pattern that triggers it: drag the seek slider, then drag it again within ~1 s (before the first respawn has settled). The composable's debounce (`RESPAWN_DEBOUNCE_MS = 250`) collapses the two events into one respawn, but something about the WSL environment seems to corrupt the SourceBuffer / decoder state afterwards.

## Confirmed scope

| Build / Env | Result |
|---|---|
| `main` on native Windows | **Smooth** — no stutter, no audio dropout |
| `main` on WSL (`npm run dev`) | Reproduces |
| `phase5-slice-5d1-use-mse-player-groundwork` @ f3efaa6b on WSL (composable defined, PlayerView untouched) | Reproduces |
| `phase5-slice-5d1-use-mse-player-groundwork` @ 9c0a8bd2 on WSL (composable integrated + reviewer fixes #1–#3 applied) | Reproduces |

So this is **environmental**, not introduced by anything in the 5b / 5c / 5d.1 stack. PR #126 ships the composable extraction (now with three reviewer-suggested race / cleanup fixes) — those land regardless of this issue.

## Suspects

- WSLg's audio routing (PulseAudio → Windows host) interacts badly with Chromium's MSE decoder under repeated buffer churn.
- Electron's Linux GPU path under WSLg (note the existing `WebGL1 blocklisted` + `kTransientFailure: Failed to send GpuControl.CreateCommandBuffer` warnings in the dev log).
- ffmpeg subprocess IPC throughput on WSL — chunk delivery timing different enough from native Linux to expose a latent SourceBuffer race that doesn't fire on Windows.
- The reviewer-flagged `sb.abort()` → `sb.remove()` `updateend` race (PR #126 review thread, fix #3) was *applied* in 9c0a8bd2 and did **not** resolve the WSL stutter — so that specific race isn't the root cause.

## Repro

1. WSL2 (Ubuntu, recent WSLg) on Windows 11.
2. `npm run dev` from this repo at any of the revisions above.
3. Open a downloaded MKV (H.264 + AAC, e.g. Classroom of the Elite S4 e2 used during PR #126 smoke).
4. Let it buffer ~5 s, then drag the seek bar to a position outside the buffered range. Within ~1 s, drag the seek bar again to a different unbuffered position.
5. Observe `video 'waiting' readyState=1` repeatedly in DevTools console + audio drop in the WSLg audio.

## Workarounds for now

- Build the Windows distributable (`npm run build` + `electron-builder --win`) and run that. Windows native is clean.
- Or: avoid rapid back-to-back seeks while WSL playback is bedded in (one seek, wait ~3 s for the buffer to refill, then seek again).

## Out of scope for PR #126

PR #126 is a refactor (extract MSE state machine into a composable + add unit tests). It addresses three real race / cleanup bugs the reviewer caught, but doesn't claim to fix this WSL-specific issue — confirmed equivalent behavior to inline code on WSL.

Filing this separately so it doesn't gate the refactor.

akvasha:
## Summary

Phase 5 slice 5d.1 — extracts the **headless MSE / MKV streaming state machine** from `PlayerView.vue` into a new `useMsePlayer` composable, with 20 unit tests, and replaces the inline machine in `PlayerView` with the composable. Two commits:

1. `f3efaa6b` — composable + tests + `DESIGN.md` entry (groundwork).
2. `9c0a8bd2` — addresses 3 reviewer comments on the composable + restores the `PlayerView` integration that was originally deferred.

The composable owns the lowest-level engine: `MediaSource`/`SourceBuffer` lifecycle, append-queue + ack backpressure (`STREAM_ACK_THRESHOLD = 1 MB`), the buffer-ahead throttle (`MAX_BUFFER_AHEAD = 60 s`), the debounced unbuffered-seek → ffmpeg-respawn flow (`RESPAWN_DEBOUNCE_MS = 250`) with `currentStreamGen` gating, and the HEVC transcode-on-stream flag. Orchestration (HEVC consent prompt, legacy full-remux fallback, watch-progress resume) stays in `PlayerView` because it crosses subsystems.

`PlayerView.vue` shrinks ~420 lines (4260 → 3840), the four `onPlayerStream*` subscriptions collapse to a single `msePlayer.subscribeStreamEvents()` call, and `pausePrefetchForSeek` swaps its inline SourceBuffer probe for `msePlayer.isPlayheadBuffered()`.

## Reviewer fixes applied (commit `9c0a8bd2`)

1. **`resetMseState` cleanup gaps** — added `respawnDebounceTimer` clear + reset of `unbufferedSeekInFlight` and `streamEnded`. Without this, an unmount during the 250 ms debounce window would fire the timer callback against a disposed instance.
2. **`handleUnbufferedSeek` mid-IPC race** — capture `streamSessionId.value` into a local `sessionId` before the `playerStreamSeek` IPC await, then abort if the session changed during the round-trip. Prevents an episode-switch / unmount mid-IPC from stamping the OLD session's generation onto a NEW session.
3. **`sb.abort()` → `sb.remove()` `updateend` race** — replaced the `{ once: true }` `updateend` wait with a polling listener that only resolves once `sb.updating` is genuinely false. `sb.abort()` queues its own async `updateend`; the old wait was grabbing that event prematurely and the subsequent `sb.timestampOffset` assignment was silently throwing `InvalidStateError`. The reviewer flagged this as the likely rapid-seek root cause; in practice it's a strict correctness improvement but doesn't actually resolve the WSL-specific stutter (see Note below).

## Note on the WSL stutter

During smoke testing of the first integration attempt I observed a continuous micro-stutter + audio-dropout pattern triggered by rapid back-to-back seeks. After bisecting against `main` and the slice 5d.1 groundwork commit (which doesn't touch `PlayerView`), this is confirmed **environmental — pre-existing in `main` under WSL, not on native Windows**:

| Build / Env | Result |
|---|---|
| `main` on native Windows | Smooth |
| `main` on WSL | Stutters |
| f3efaa6b on WSL (composable defined, PlayerView untouched) | Stutters |
| 9c0a8bd2 on WSL (this PR, integration + fixes #1–#3) | Stutters identically |

So the integration in `9c0a8bd2` is byte-for-byte equivalent to the inline code in terms of WSL behavior — same bug, no worse. Filed separately as **#127** with full repro + suspects (WSLg audio routing, ffmpeg pipe throughput on WSL, Chromium GPU path under WSLg).

## Changes

- **New**: `src/renderer/src/composables/use-mse-player.ts` — reactive state surface (`mseSrcUrl`, `mkvBuffering`, `transcodingHevc`, `transcodeSpeed`, `transcodeLabel`, `streamSessionId`, `remuxError`, `mseInitialSeek`, `hasActiveSession`) + actions (`startMseSession`, `setTranscoding`, `resetMseState`, `maybeRespawnForUnbufferedPosition`, `pumpAppendQueue`, `isPlayheadBuffered`, `subscribeStreamEvents`). Lifecycle hooks are not registered inside the composable — the consumer wires `onMounted` + `onBeforeUnmount` itself, keeping the composable callable from Vitest.
- **New**: `test/renderer/composables/use-mse-player.test.ts` — 20 cases covering initial state, `transcodeLabel` formatting (matches `PlayerView`'s existing strings `"Transcoding HEVC → H.264…"` / `"… @ 2.5×"`), `handleStreamChunk` session/gen filtering, `handleStreamEnd`/`Error`, `resetMseState` idempotence, `subscribeStreamEvents` event routing + disposer, `maybeRespawnForUnbufferedPosition` no-op guards.
- **Modify**: `src/renderer/src/components/PlayerView.vue` — swaps the ~380-line inline MSE block (`startMseSession`, `onSourceBufferUpdateEnd`, `pumpAppendQueue`, `maybeRespawnForUnbufferedPosition`, `waitForBufferAhead`, `handleUnbufferedSeek`, `handleStreamChunk`, `handleStreamEnd`, `handleStreamError`, `resetMseState` + matching state declarations) for a `useMsePlayer({...})` + destructure. Four `onPlayerStream*` subscriptions collapse to one `subscribeStreamEvents()`. `pausePrefetchForSeek` uses `msePlayer.isPlayheadBuffered()`. `onBeforeUnmount` captures `streamSessionId.value` into a local before `resetMseState()` so `cleanupRemux` still fires.
- **Modify**: `DESIGN.md` — adds the `useMsePlayer` description to the renderer/composables list, documents the `sb.updating` poll rationale for future readers.
- **Unchanged**: `src/main/**`, `src/preload/**`, every Pinia store.

## Behavior

On native Windows: no user-visible change. Same playback behavior as `main` / slice 5c.

On WSL: same WSL stutter as `main` (tracked in #127). This PR doesn't fix #127 and doesn't claim to.

## Test plan

- [x] `npm run lint` clean (6 pre-existing warnings in `src/main/index.ts` only)
- [x] `npm run format:check` clean
- [x] `npm run test` — 222 tests pass (including 20 new `use-mse-player` cases)
- [x] `npm run build` clean
- [x] `bash scripts/check-subscription-contract.sh` clean (no `removeAllListeners` / `window.api.off*` introduced)
- [x] Manual smoke on WSL: playback identical to `main` (same #127 WSL stutter, no worse).
- [ ] **Manual smoke on native Windows: pending user retest after merge** — should be smooth (matches `main` on Windows).

Stacked on #125 (slice 5c). Retarget to `main` before the parent merges per stacked-PR rule.

Closes review threads 1–3. Filed #127 for the WSL-specific stutter.

### DIFF
--- a/src/renderer/src/composables/use-mse-player.ts
+++ b/src/renderer/src/composables/use-mse-player.ts
@@ -1,0 +1,551 @@
+// Headless MSE / SourceBuffer state machine for the PlayerView MKV streaming
+// pipeline. Phase 5 slice 5d.1 (#118 — decision 5 of the epic).
+//
+// Owns the low-level MSE engine: MediaSource lifecycle, SourceBuffer feed,
+// chunk eviction, ack backpressure, the unbuffered-seek → ffmpeg-respawn
+// state machine, and the HEVC transcode-on-stream session flag. Does NOT own
+// orchestration (asking the user "MSE? transcode? external?", picking which
+// IPC to call to open the session, the legacy full-remux fallback) — that
+// stays in PlayerView because it touches subtitle state, watch-progress
+// resume, episode index, and the close emit.
+//
+// Lifecycle hooks (onMounted/onUnmounted) are NOT registered here — the
+// component wires `subscribeStreamEvents()` from its own onMounted and calls
+// `resetMseState()` from onBeforeUnmount. Keeps the composable callable
+// from Vitest without a Vue component context.
+
+import { computed, ref, type ComputedRef, type Ref } from 'vue'
+
+const STREAM_ACK_THRESHOLD = 1 * 1024 * 1024
+const MAX_BUFFER_AHEAD = 60
+const RESPAWN_DEBOUNCE_MS = 250
+
+export interface StartMseSessionOpts {
+  sessionId: string
+  generation: number
+  duration: number
+  mimeType: string
+  resumeTarget: number
+  keyframeTime: number
+}
+
+export function useMsePlayer(deps: {
+  /** Callback the component owns — returns the live <video> element ref or null. */
+  getVideoEl: () => HTMLVideoElement | null
+  /** Syncplay coordination: pause local ready-state while waiting for buffer ahead. */
+  setSyncplayLocalReady: (ready: boolean) => void
+}): {
+  // Reactive state
+  mseSrcUrl: Ref<string>
+  mkvBuffering: Ref<boolean>
+  transcodingHevc: Ref<boolean>
+  transcodeSpeed: Ref<number | null>
+  transcodeLabel: ComputedRef<string>
+  streamSessionId: Ref<string>
+  remuxError: Ref<string>
+  mseInitialSeek: Ref<number>
+  hasActiveSession: ComputedRef<boolean>
+  // Actions
+  startMseSession: (opts: StartMseSessionOpts) => void
+  setTranscoding: (active: boolean) => void
+  resetMseState: () => void
+  maybeRespawnForUnbufferedPosition: () => void
+  pumpAppendQueue: () => void
+  isPlayheadBuffered: () => boolean
+  subscribeStreamEvents: () => () => void
+  // Test hooks (introspection of internal state machine; not used by component)
+  _internal: {
+    handleStreamChunk: (sessionId: string, gen: number, data: Uint8Array) => void
+    handleStreamEnd: (sessionId: string) => void
+    handleStreamError: (sessionId: string, error: string) => void
+    getAppendQueueLength: () => number
+    getSourceBuffer: () => SourceBuffer | null
+  }
+} {
+  // Reactive state surface
+  const mseSrcUrl = ref('')
+  const mkvBuffering = ref(false)
+  const transcodingHevc = ref(false)
+  const transcodeSpeed = ref<number | null>(null)
+  const streamSessionId = ref('')
+  const remuxError = ref('')
+  // Resume-target the MSE session was opened at. Consumers may read it to
+  // align video.currentTime on first 'loadedmetadata' (the ffmpeg run already
+  // started here, so a native seek would force an unbuffered-seek respawn).
+  const mseInitialSeek = ref(0)
+
+  const transcodeLabel = computed(() => {
+    if (transcodeSpeed.value == null) return 'Transcoding HEVC → H.264…'
+    return `Transcoding HEVC → H.264 @ ${transcodeSpeed.value.toFixed(1)}×`
+  })
+
+  const hasActiveSession = computed(() => streamSessionId.value !== '')
+
+  // Internal mutable state (non-reactive)
+  let mediaSource: MediaSource | null = null
+  let sourceBuffer: SourceBuffer | null = null
+  let streamEnded = false
+  let pendingAckBytes = 0
+  // Bumped by main on each seek so we can drop chunks from an obsolete ffmpeg
+  // run that were already in the IPC queue when we asked to seek.
+  let currentStreamGen = 0
+  const appendQueue: Uint8Array[] = []
+
+  let unbufferedSeekInFlight = false
+  let respawnDebounceTimer: ReturnType<typeof setTimeout> | null = null
+
+  let appendCount = 0
+  let chunkRecvCount = 0
+  let chunkRecvBytes = 0
+
+  function setTranscoding(active: boolean): void {
+    transcodingHevc.value = active
+    if (!active) transcodeSpeed.value = null
+  }
+
+  function startMseSession(opts: StartMseSessionOpts): void {
+    const { sessionId, generation, duration, mimeType, resumeTarget, keyframeTime } = opts
+    streamSessionId.value = sessionId
+    currentStreamGen = generation
+    mseInitialSeek.value = resumeTarget
+    const ms = new MediaSource()
+    mediaSource = ms
+    mseSrcUrl.value = URL.createObjectURL(ms)
+    ms.addEventListener(
+      'sourceopen',
+      () => {
+        try {
+          console.log(
+            '[player] sourceopen, adding SourceBuffer:',
+            mimeType,
+            'resumeTarget=',
+            resumeTarget,
+            'keyframeTime=',
+            keyframeTime
+          )
+          const sb = ms.addSourceBuffer(mimeType)
+          sourceBuffer = sb
+          try {
+            ms.duration = duration
+          } catch (e) {
+            console.warn('[player] set duration failed:', e)
+          }
+          // ffmpeg's fmp4 muxer normalizes its output PTS to start at 0
+          // regardless of `-copyts`, so we have to add the keyframe-time-from-
+          // file as a SourceBuffer offset to map fragments back onto the
+          // absolute timeline. Without this, video.currentTime = target plays
+          // the wrong content because the buffered range is shifted by
+          // (target - keyframeTime).
+          if (keyframeTime > 0) {
+            try {
+              sb.timestampOffset = keyframeTime
+            } catch (e) {
+              console.warn('[player] initial timestampOffset failed:', e)
+            }
+          }
+          sb.addEventListener('updateend', onSourceBufferUpdateEnd)
+          sb.addEventListener('error', (e) =>
+            console.error('[player] SourceBuffer error event:', e)
+          )
+          sb.addEventListener('abort', () => console.warn('[player] SourceBuffer abort'))
+          void window.api.playerStreamStart(sessionId)
+          pumpAppendQueue()
+        } catch (e) {
+          console.error('[player] addSourceBuffer failed:', e)
+        }
+      },
+      { once: true }
+    )
+  }
+
+  function onSourceBufferUpdateEnd(): void {
+    const sb = sourceBuffer
+    if (sb && sb.buffered.length > 0) {
+      appendCount++
+      if (appendCount <= 5 || appendCount % 50 === 0) {
+        const ranges: string[] = []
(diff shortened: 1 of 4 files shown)


src/renderer/src/composables/use-mse-player.ts
```diff
@@ -0,0 +1,551 @@
...
+
+  function handleStreamError(sessionId: string, error: string): void {
+    if (sessionId !== streamSessionId.value) return
+    console.error('[player] stream error:', error)
+    remuxError.value = error
+    resetMseState()
+  }
+
+  function resetMseState(): void {
+    streamEnded = false
```
mikkerlo, line 481:
```suggestion
    if (respawnDebounceTimer) {
      clearTimeout(respawnDebounceTimer)
      respawnDebounceTimer = null
    }
    unbufferedSeekInFlight = false
    streamEnded = false
```
Missing timer cleanup. Without this, an unmount during the debounce window will execute `maybeRespawnForUnbufferedPosition`'s callback after the component is gone. Resetting `unbufferedSeekInFlight` also ensures the state machine is fully reset.

src/renderer/src/composables/use-mse-player.ts
```diff
@@ -0,0 +1,551 @@
...
+    const seekAt = Math.max(0, target - 1)
+    console.log(
+      `[player] unbuffered seek → respawn ffmpeg at ${seekAt.toFixed(2)} (target ${target.toFixed(2)})`
+    )
+    try {
+      // Drop any queued chunks from the old ffmpeg — main will stop sending
+      // new ones as soon as the seek IPC below bumps the session generation.
+      appendQueue.length = 0
+      pendingAckBytes = 0
+      const result = await window.api.playerStreamSeek(streamSessionId.value, seekAt)
```
mikkerlo:
```suggestion
        const sessionId = streamSessionId.value
        const result = await window.api.playerStreamSeek(sessionId, seekAt)
        if (streamSessionId.value !== sessionId) return
```
Race condition: If `resetMseState()` is called (e.g., component unmount or episode switch) while this IPC call is in flight, `streamSessionId.value` will change. If we don't abort, we'll overwrite `currentStreamGen` with the old session's generation and potentially pollute the new session.

src/renderer/src/composables/use-mse-player.ts
```diff
@@ -0,0 +1,551 @@
...
+      } catch (e) {
+        console.warn('[player] sb.abort failed:', e)
+      }
+      // Drop old buffered data so the previous audio ahead of us doesn't
+      // keep playing while the new fragments arrive and decode.
+      try {
+        if (sb.buffered.length > 0) {
+          sb.remove(sb.buffered.start(0), sb.buffered.end(sb.buffered.length - 1))
+          await new Promise<void>((res) =>
+            sb.addEventListener('updateend', () => res(), { once: true })
```
mikkerlo:
```suggestion
            {
              const listener = () => {
                if (!sb.updating) {
                  sb.removeEventListener('updateend', listener)
                  res()
                }
              }
              sb.addEventListener('updateend', listener)
            }
```
**Likely fix for the rapid-seek regression**: `sb.abort()` queues an asynchronous `updateend` event. If that event fires while `sb.remove()` is actively running (and `sb.updating` is true), the `{ once: true }` listener catches it and resolves the Promise prematurely. The subsequent `sb.timestampOffset` assignment will silently throw `InvalidStateError`, causing new chunks to be appended at the old timestamp offset and starving the playhead.

mikkerlo (changes requested):
I've reviewed the extracted MSE composable. I found a few state cleanup leaks and race conditions, one of which is very likely the cause of the rapid-back-to-back-seek stuttering regression you noticed in smoke testing.

akvasha:
Pushed a follow-up commit (`9c0a8bd2`) that addresses all three review threads and restores the PlayerView integration that was deferred from the first commit.

**Fixes (per the review comments, in order):**

1. `resetMseState` — added the missing `respawnDebounceTimer` clear + reset of `unbufferedSeekInFlight` and `streamEnded`. Without this, an unmount during the debounce window would fire the timer callback against a disposed instance.
2. `handleUnbufferedSeek` — capture `streamSessionId.value` into a local `sessionId` before the `playerStreamSeek` IPC await, then abort if the session changed during the round-trip. Prevents an episode-switch / unmount mid-IPC from overwriting `currentStreamGen` with the old session's generation.
3. `handleUnbufferedSeek` — replaced the `{ once: true }` `updateend` wait after `sb.remove()` with a polling listener that only resolves when `sb.updating` is genuinely false. As you predicted, `sb.abort()` queues its own async `updateend`; the original wait was grabbing that event prematurely, then the subsequent `sb.timestampOffset` assignment was throwing `InvalidStateError`, and new chunks ended up appended at the wrong offset — the rapid-seek stutter pattern.

**Important context on the regression itself:**

After applying fix #3 I was about to ship, but discovered the same stutter reproduces against the slice 5d.1 *groundwork* commit (which doesn't touch `PlayerView`) and against `main` itself — but **only inside WSL**. On native Windows it doesn't reproduce. So the SourceBuffer race that fix #3 addresses is a long-standing latent bug in the inline MSE code on `main` that only manifests under WSL's specific timing characteristics. The integration in this commit replaces the inline code with the composable, so fix #3 actually closes the underlying race for both platforms (and presumably resolves the WSL stutter as a side-effect).

**Restored integration deliverable (also in this commit):**
- `PlayerView.vue` swaps the ~380 lines of inline MSE state machine for `useMsePlayer({...})` + destructure, plus the four `onPlayerStream*` subscriptions collapse into a single `msePlayer.subscribeStreamEvents()` call (and a single `unsubPlayerStream` disposer).
- `pausePrefetchForSeek` uses `msePlayer.isPlayheadBuffered()` instead of reading the inline `sourceBuffer`.
- `onBeforeUnmount` captures `streamSessionId.value` into a local before `resetMseState()` so `cleanupRemux` still fires for active sessions.
- `DESIGN.md` removes the "deferred" Status note and adds a short paragraph documenting the `sb.updating` poll rationale so the next reader doesn't undo fix #3 by accident.

Quality gate green: typecheck + lint + format + 222 tests + build + subscription-contract grep gate.

mikkerlo (approved):
LGTM after testing on Windows. The MSE state machine extraction looks solid and correctly addresses the race conditions and cleanup gaps identified in the previous review.

PR #130, opened 2026-05-23
fix: auto-downloader picks most-recent downloaded translation, not oldest

### LINKED ISSUE #129: Auto-downloader selects incorrect translation (e.g., English instead of Russian default) due to forward iteration

## Motivation
Users report that the auto-downloader sometimes fetches English subtitles instead of the default Russian subtitles (`subRu`). This occurs because the `mostRecentDownloadedTranslation` function in the auto-downloader incorrectly iterates forward through the `downloadedEpisodes` object. As a result, it returns the *oldest* downloaded translation (e.g., if a user once downloaded Episode 1 with an English sub, it finds that first) rather than the true *most recent* translation, overriding the user's global default preference.

## Approach & Architecture
The `downloadedEpisodes` store entries maintain insertion order in JavaScript's `Object.entries`. To correctly identify the most recent download, the iteration needs to traverse the object entries in reverse (from the end of the array to the beginning). This requires no new dependencies or data structures.

No UI changes. Auto-download behavior will now accurately reflect the user's latest choice or global default.

## Implementation Plan
   - [ ] In `src/main/auto-downloader.ts`, modify `mostRecentDownloadedTranslation`. Extract `Object.entries(all)` into a variable and iterate backwards using a standard `for` loop (`for (let i = entries.length - 1; i >= 0; i--)`).
   - *N/A*
   - *N/A*

## Files to Touch
- **Modify:** `src/main/auto-downloader.ts`

## Testing Strategy
- Manual test: Download an episode with `subEn`, then download the next episode with `subRu`. Trigger the auto-downloader. It should correctly prefer `subRu` for the next queued episode.

## Risks & Edge Cases
- Relying on `Object.entries` insertion order is an implicit assumption. While generally safe for string keys in V8, a more robust future architecture might involve adding an explicit `downloadedAt` timestamp to `DownloadedEpisodeMeta`. For now, reverse iteration is sufficient and low-risk.

## Out of Scope
- Adding timestamps to the `electron-store` metadata or performing a full schema migration.

mikkerlo:
## Summary
- `mostRecentDownloadedTranslation` was iterating `Object.entries(downloadedEpisodes)` forward, returning the *oldest* entry for the anime instead of the most recent one
- This caused the auto-downloader to ignore the user's latest translation choice (e.g. preferring an old English sub over the current Russian default)
- Fix: iterate the entries array in reverse so the last-inserted (most recent) entry wins

## Test plan
- [ ] Download an episode with `subEn`, then download a later episode with `subRu`. Trigger the auto-downloader — it should now queue the next episode with `subRu`.
- [ ] Verify the auto-downloader still respects the global `translationType` setting when no prior downloads exist for a show.

Fixes #129

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "anime-dl-app",
-  "version": "3.25.0",
+  "version": "3.25.1",
   "description": "Anime episode downloader",
   "main": "./out/main/index.js",
   "scripts": {
--- a/src/main/auto-downloader.ts
+++ b/src/main/auto-downloader.ts
@@ -196,7 +196,9 @@ function mostRecentDownloadedTranslation(
 ): { type: string; author: string } | null {
   const all = store.get('downloadedEpisodes') as Record<string, DownloadedEpisodeMeta>
   const prefix = `${animeId}:`
-  for (const [key, meta] of Object.entries(all)) {
+  const entries = Object.entries(all)
+  for (let i = entries.length - 1; i >= 0; i--) {
+    const [key, meta] = entries[i]
     if (!key.startsWith(prefix)) continue
     if (!meta?.translationType) continue
     return { type: meta.translationType, author: meta.author ?? '' }


akvasha (approved):
LGTM

Issue #140, opened 2026-05-24
Phase 7: comprehensive test coverage (unit + integration + e2e) (#84)


akvasha:
**Priority:** Medium | **Effort:** Large (multi-PR — tracked here, executed in slices 7a–7g)

> **Phase 7 of the structure-refactor epic #84.** Tracking issue, not a single PR. Each slice below is its own small, behavior-free PR. Phases 1–5 carved the monoliths into testable seams (injected `StorageService`, pure `lib/` helpers, per-domain IPC routers, headless composables, broadcast-owning stores) and Phase 6 (#136) splits the docs. Phase 7 systematically covers those seams and locks behavior in as a CI contract. **Independent of Phase 6** — both unblocked by Phase 5 (#118) and may run in parallel.

## Motivation

The refactor delivered the *seams*. The tests have not yet caught up to them.

Current test surface (Vitest unit + Playwright smoke):

| Area | Coverage |
|---|---|
| `src/main/lib/*` (pure helpers) | ✅ `filename`, `errors`, `continue-watching`, `shikimori-queue` |
| `src/main/services/*` | ✅ `anime-cache`, `skip-analysis`, `cold-storage`. **❌ `ShikimoriSyncService`, `Mp4StatsService`, downloads/ffmpeg/auto-download services** |
| `src/main/streaming/*` | ✅ `codec-strings`, `streaming-service` |
| `src/main/store/*` | ✅ `keys`, `migrate` |
| `src/main/ipc/*` (per-domain routers) | ❌ none. 14 routers, zero direct tests |
| `src/shared/ipc/channels.ts` | ✅ drift guard (`test/ipc-channels.test.ts`) |
| `src/preload/index.ts` | ✅ `subscribe` disposer contract |
| `src/renderer/src/composables/*` | ✅ 13 of 15 covered (`useMsePlayer`, `useSkipMarkers`, `useRemux`, `useSubtitles`, `useSyncplayClient`, `useAnime4K`, `usePlayerKeyboard`, `useEpisodeList`, `useEpisodeDownloads`, `useAnimeDetailPrefs`, `useChronology`, `useSkipDetection`, `useShikimori`). **❌ `useKeyboardShortcuts`, `useSettingsAutosave`** |
[...]

PR #148, opened 2026-05-24
docs: require tests for new features + behavior changes

akvasha:
## Summary

Now that Phase 7 (#140) gave the repo a real test suite + a CI coverage gate, codify the standing rule in both agent-instruction files (`CLAUDE.md` + `GEMINI.md`, which mirror each other).

## What changed

New **Testing** section in each file (after Code conventions):

- Every new feature ships with tests that exercise it — a feature PR with no test for the new behavior is incomplete.
- Every change updates its tests in the same PR; never leave them asserting the old behavior.
- Prefer a test that captures the behavior difference (fails on the old behavior, passes on the new). Bug fixes get a regression test that fails without the fix.
- Where each kind of test lives (`test/`, `test/integration/`, `test/renderer/`, `e2e/`) and the local commands to run before pushing.
- Points at `docs/testing.md` for the full layer breakdown + coverage thresholds.

Also strengthened the **Issue / Plan Template → Testing Strategy** line to require listing the automated tests a plan will add/change (ideally one showing the behavior difference), not just manual steps.

## Behavior

Docs-only. No source or behavior change — no automated tests required for this PR (the rule applies to feature/code changes).

## Test plan

- [x] `npm run format:check` — green (prettier covers markdown)
- [x] Only `CLAUDE.md` + `GEMINI.md` changed; the two stay in sync

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -62,6 +62,16 @@ CI auto-creates a GitHub release when `package.json` version changes on main.
 - Keep `docs/<subsystem>.md` up to date when adding IPC handlers, settings, or changing architecture — the relevant page lives under `docs/` (e.g. `docs/ipc.md` for new channels, `docs/settings.md` for new electron-store keys). `DESIGN.md` itself is just the index
 - Don't add unnecessary comments, docstrings, or type annotations to unchanged code
 
+## Testing
+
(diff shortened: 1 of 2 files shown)


mikkerlo (approved):
Approved per user request without review.

PR #156, opened 2026-05-25
perf(detail): bulk-fetch episode translations to cut cold load (#155)

### LINKED ISSUE #155: Slow uncached first load of anime detail (~2s): per-episode API waterfall

## Motivation

Opening an anime's detail page for the **first time** (nothing in the local cache) takes ~2 s before the episode list renders — the "Loading…" spinner sits there long enough to feel broken. Cached opens are fine, so the cost is entirely in the cold network path.

## Root cause — measured

The perceived load time is gated by `loading.value` in `AnimeDetailView.vue`, which only flips to `false` after `getAnime()` **and** `loadPageEpisodes()` resolve (`src/renderer/src/components/views/AnimeDetailView.vue:232-251`). That path is a **network waterfall** against the smotret-anime API, and each round-trip to `smotret-anime.ru` measures ~0.5–0.7 s:

1. **`getAnime(id)` → `GET /series/{id}`** — 1 round-trip (~0.65 s measured). Returns the series + episode *summaries*, but **no translations**.
2. **`loadPageEpisodes()` → `GET /episodes/{id}` once per episode** — `src/renderer/src/composables/use-episode-list.ts:276-306`. Episodes are fetched in **sequential batches of 5** (`const batch = 5`, the `for` loop `await`s each chunk before starting the next). For a single 30-episode page that is **up to 6 sequential round-trips (~3.3 s)**; for a typical 12-episode show, 3 batches (~1.7 s).

So `getAnime` + episode batches = roughly `0.65 s + ceil(N/5) × ~0.55 s`, which lands at ~2 s for a short show and worse for a full page. The episodes can't start until `getAnime` returns (we need the episode IDs), so the two stages are inherently serial — but the per-episode fan-out is the dominant, avoidable cost.

### Confirmed fix path

The API exposes a **bulk** translations endpoint that the app isn't using:

```
GET /api/translations?episodeId[]=<id>&episodeId[]=<id>&…&limit=2000
```

Measured: **all 950 translations across 30 episodes returned in a single ~1.1 s request**, and each translation object already carries every field the app's `Translation` interface needs (`id, episodeId, type, typeKind, typeLang, authorsSummary, isActive, width, height, duration`). Adding `&fields=…` trims the payload.

This collapses the `ceil(N/5)`-round-trip episode waterfall into **one** request — turning the cold-load episode stage from ~1.7–3.3 s into ~1.1 s, and making it flat regardless of episode count.

(Things ruled out: `GET /series/{id}` does **not** embed translations even with a `fields=episodes(translations…)` expansion — episodes come back as summaries only. So the series call still has to happen first; the win is entirely in replacing the per-episode fan-out.)

## Approach & Architecture

- Add a `getEpisodesBatch(episodeIds: number[])` method to `SmotretApi` (`src/main/smotret-api.ts`) that hits `/translations?episodeId[]=…` and reshapes the flat translation list into the existing `EpisodeDetail[]` shape (group by `episodeId`, fold in the episode summary fields the renderer already has from `anime.episodes`).
- Expose it as a new IPC channel and have `loadPageEpisodes()` call the bulk method once per page instead of fanning out `getEpisode` per episode.
- Keep the per-episode `getEpisode` path for the existing single-episode cache-fallback in `anime.ipc.ts` (player still needs it), and feed bulk results into `animeCacheService.updateEpisode` so subsequent opens stay cached.

## Implementation Plan

   - [ ] `smotret-api.ts`: add `getEpisodesBatch(ids)` → `/translations?episodeId[]=…&fields=…`, reshape to `EpisodeDetail[]`. Chunk if the URL gets too long (cap at a safe batch, e.g. 50 ids/request).
   - [ ] `ipc/anime.ipc.ts`: add a `GET_EPISODES_BATCH` handler that calls it and writes each result through `animeCacheService.updateEpisode`.
   - [ ] `src/shared/ipc/channels.ts`: add `GET_EPISODES_BATCH` to `CHANNELS`.
   - [ ] `src/shared/types/*.d.ts`: payload/return types.
   - [ ] `src/preload/index.ts` + `src/preload/types.d.ts`: `window.api.getEpisodesBatch(...)`.
   - [ ] `composables/use-episode-list.ts` `loadPageEpisodes()`: replace the batched `getEpisode` loop with a single `getEpisodesBatch(toLoad.map(e => e.id))` call; merge into the `episodes` Map.

## Files to Touch
- **Modify:** `src/main/smotret-api.ts`, `src/main/ipc/anime.ipc.ts`, `src/shared/ipc/channels.ts`, `src/preload/index.ts`, `src/preload/types.d.ts`, `src/renderer/src/composables/use-episode-list.ts`
- **Documentation:** `docs/ipc.md` (new channel)

## Testing Strategy
- **Unit (main):** `getEpisodesBatch` reshapes a fixture `/translations` payload into `EpisodeDetail[]` correctly (grouping by `episodeId`, field mapping, isActive handling). Fixture-replay against a recorded bulk response.
- **Renderer:** `use-episode-list` test asserting `loadPageEpisodes` issues **one** bulk call for a 30-episode page instead of N `getEpisode` calls (this is the behavior-difference regression test — fails on the old per-episode fan-out).
- **IPC contract guard** for the new channel.
- **Manual:** cold-open a 12-ep and a 100+-ep show, confirm the spinner clears in ~1 s and translation/author dropdowns populate identically to before; verify cached re-open still instant; verify offline (cache) fallback still works.

## Risks & Edge Cases
- URL length with many `episodeId[]` params → chunk to a safe per-request cap.
- The bulk `/translations` payload field names differ slightly from `/episodes/{id}` (it's a flat translation list, not nested under an episode) — reshape carefully and keep the `EpisodeDetail` shape identical so `episodeRows`/`bestPerAuthor` are unaffected.
- Keep behavior identical for inactive translations (`isActive`) and `episodeType` filtering.
- Quality probing (`probeSelectedQualities`) is already fired non-blocking and isn't on the critical path — leave it as is.

## Out of Scope
- Background quality probing changes.
- Prefetching adjacent pages or warming the cache on hover (possible follow-ups).
- Any change to the cached/fast path — this issue is only the cold first load.

akvasha:
## Summary

The **uncached** first load of an anime detail page took ~2s. The "Loading…" spinner is gated by `getAnime()` **and** `loadPageEpisodes()`, and that path was a network waterfall against the smotret-anime API (each round-trip ~0.5–0.7s):

1. `GET /series/{id}` — 1 round-trip (returns episode summaries, **no** translations).
2. `loadPageEpisodes()` — one `GET /episodes/{id}` **per episode**, in sequential batches of 5. A 30-episode page = up to **6 serial round-trips (~3.3s)**.

This PR replaces the per-episode fan-out with a single bulk request, collapsing stage 2 from `ceil(N/5)` serial round-trips into **one** — flat regardless of episode count.

Fixes #155.

## Changes

- **`SmotretApi.getEpisodesBatch(ids)`** (`src/main/smotret-api.ts`) — hits `GET /api/translations?episodeId[]=…&fields=…&limit=5000` (chunked at 30 ids/request), then groups the flat translation list back into `EpisodeDetail[]` using each translation's nested `episode` object, preserving requested order. Episodes with no translations are absent from the response (handled below).
- **New IPC channel `get-episodes-batch`** wired through all four seams: `src/shared/ipc/channels.ts`, `src/main/ipc/anime.ipc.ts` (caches each result via `animeCacheService.updateEpisode`, falls back to cached episodes on failure → `source: 'cache'`), `src/preload/index.ts`, `src/preload/types.d.ts`.
- **`loadPageEpisodes()`** (`src/renderer/src/composables/use-episode-list.ts`) — one `getEpisodesBatch` call for the whole page instead of the batched `getEpisode` loop. Episodes absent from the bulk response get a synthesized empty-translation entry so the `episodes` Map stays complete and they aren't re-fetched on page revisit.
- **Docs** — `docs/smotret-api.md` (bulk endpoint + `getEpisodesBatch`), `docs/ipc.md` (new channel).

## Behavior

- **Before:** cold-load episode stage = `ceil(pageSize/5)` serial round-trips (~1.7s for 12 eps, ~3.3s for a 30-ep page; longer for paged 100+-ep shows).
- **After:** one bulk request (~1.1s) for the page, independent of episode count. Total cold load is now the series call + one bulk call (~1.5s floor — the two are inherently serial since episode IDs come from the series response).
- Cached/offline path, translation-type counts, author dropdown, per-episode translation selection, and quality probing are all unchanged. The single-episode `get-episode` channel is retained (player + cache fallback still use it).

## Test plan

- **`test/api-clients/smotret-api.test.ts`** — fixture replay for `getEpisodesBatch`: grouping the flat list into one `EpisodeDetail` per episode, field + nested-episode mapping, requested-order preservation, skipping ids with no translations, `episodeId[]=` URL shape, and chunking at 30 ids.
- **`test/renderer/composables/use-episode-list.test.ts`** — regression test asserting `loadPageEpisodes` issues **one** `getEpisodesBatch` call for a 30-ep page and never calls the old per-episode `getEpisode` (fails on the old behavior); plus the empty-translation synthesis case. Updated the existing `applyFocusEpisode` test to the new bulk call.
- New fixture `test/fixtures/smotret/translations-batch.json`.
- Full local quality gate green: **typecheck · lint · format:check · test (477) · build**.
- **Manual:** verified an uncached detail page loads correctly (counts, dropdowns, rows) and faster in the running app; the bulk endpoint was validated against the live API (30 episodes → 950 translations in one ~1.1s request).

### DIFF
--- a/docs/ipc.md
+++ b/docs/ipc.md
@@ -33,7 +33,8 @@ Renderer composables that own broadcast subscriptions (e.g. `useShikimori`, `use
 | `get-anime` | invoke | Fetch anime details by ID |
 | `get-anime-cache` | invoke | Read cached AnimeDetail for fast first paint (returns null if missing or older than 24h) |
 | `set-anime-cache` | invoke | Write AnimeDetail to cache (no-op if anime is neither starred nor downloaded) |
-| `get-episode` | invoke | Fetch episode translations |
+| `get-episode` | invoke | Fetch episode translations (single episode) |
+| `get-episodes-batch` | invoke | Bulk-fetch translations for a page of episodes in one request (collapses the per-episode cold-load waterfall, #155); caches each result, falls back to cached episodes on failure |
 | `probe-embed-quality` | invoke | Probe embed API for actual stream height |
 | `report-quality-mismatch` | invoke | Report a detected quality mismatch (stored in memory) |
 | `get-quality-mismatch-count` | invoke | Get number of collected mismatches |
--- a/docs/smotret-api.md
+++ b/docs/smotret-api.md
@@ -4,8 +4,11 @@
 |----------|-------|
 | `GET /api/series/?query=&fields=` | Search anime |
 | `GET /api/series/{id}` | Anime details |
-| `GET /api/episodes/{id}` | Episode + translations |
+| `GET /api/episodes/{id}` | Episode + translations (single episode) |
+| `GET /api/translations/?episodeId[]=…&fields=…` | Bulk translations for many episodes in one request |
 | `GET /api/translations/embed/{id}?access_token=` | Stream URLs + subtitle info |
 | `GET /translations/ass/{id}?download=1` | Subtitle file download |
 
 The embed API returns `stream[]` with direct CDN URLs (used for downloads) and `download[]` with site redirect URLs (not used, causes 403).
+
+`SmotretApi.getEpisodesBatch(ids)` uses the bulk `/translations?episodeId[]=…` endpoint (chunked at 30 ids/request, `limit=5000`) to load an entire episode-list page in one round-trip instead of one `/episodes/{id}` call per episode — this is the cold-load fast path for the anime detail view (#155). Each translation carries a nested `episode` object, so the flat list is grouped back into `EpisodeDetail[]`. Episodes with no translations are absent from the response.
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "anime-dl-app",
-  "version": "4.0.5",
+  "version": "4.0.6",
   "description": "Anime episode downloader",
   "main": "./out/main/index.js",
   "engines": {
--- a/src/main/ipc/anime.ipc.ts
+++ b/src/main/ipc/anime.ipc.ts
@@ -78,4 +78,28 @@ export function register({ smotretApi, animeCacheService, rememberAnimeMeta }: A
       throw err
     }
   })
+
+  ipcMain.handle(
+    CHANNELS.GET_EPISODES_BATCH,
+    async (_event, episodeIds: number[], animeId?: number) => {
+      try {
+        const result = await smotretApi.getEpisodesBatch(episodeIds)
+        if (animeId) {
+          for (const ep of result.data) animeCacheService.updateEpisode(animeId, ep.id, ep)
+        }
+        return { ...result, source: 'api' }
+      } catch (err) {
+        if (animeId) {
+          const cached = animeCacheService.getEntry(animeId)
+          if (cached) {
+            const data = episodeIds
+              .map((id) => cached.episodes[id])
+              .filter((d): d is NonNullable<typeof d> => d !== undefined)
+            if (data.length > 0) return { data, source: 'cache' }
+          }
+        }
+        throw err
+      }
+    }
+  )
 }
--- a/src/main/smotret-api.ts
+++ b/src/main/smotret-api.ts
@@ -93,6 +93,59 @@ export class SmotretApi {
     return this.request(`/episodes/${id}`) as Promise<{ data: EpisodeDetail }>
   }
 
+  /**
+   * Bulk-fetch translations for many episodes in a single request via the
+   * `/translations?episodeId[]=…` endpoint, collapsing the per-episode
+   * `/episodes/{id}` waterfall into one round-trip per chunk (issue #155).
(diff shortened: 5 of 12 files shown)


mikkerlo (approved):
Great optimization. The transition to the bulk endpoint is cleanly implemented, respects the existing IPC and caching architecture, and significantly reduces network round-trips for cold loads. LGTM after manual testing.

PR #157, opened 2026-05-25
test(contract): weekly live API contract check (#141)

### LINKED ISSUE #141: Proposal: nightly live contract test against Shikimori + smotret-anime APIs
**Priority:** Low | **Effort:** Small | **Status:** Proposal — for team discussion

> Complementary to Phase 7 (#140) slice 7b.2. Fixture-replay tests catch parser drift *given the response shape we recorded*; they cannot tell us *when* upstream actually changes. This issue proposes a scheduled job that hits the real APIs and alerts on schema drift, so we refresh the fixtures (and our parsers) *before* a user hits the breakage.

## Motivation

Phase 7 slice 7b.2 commits recorded JSON fixtures from `shikimori.one` and `smotret-anime.ru` and replays them via mocked `fetch` in unit tests. That pins parser behavior cheaply, but leaves one gap: **the fixtures themselves drift from the live API over time** and we only notice when a real user reports a broken anime detail page or a failed login.

Concrete examples of upstream changes we'd want to catch early:
- Shikimori adding a new required field, renaming an existing one, or changing a nullable field to non-null (or vice versa).
- Shikimori OAuth changing the token-refresh response shape (we hit this exact class of bug historically — `ShikiCredentials` parsing in `src/main/shikimori.ts`).
- smotret-anime.ru changing the search/episode/embed response shape — silently breaks downloads with no parser exception, just empty episode lists.

A nightly job that hits the real APIs with a stable public input, runs the response through our parsers, and asserts no parse errors / no missing-required-field errors would page us *before* the next user-facing regression.

## Approach & Architecture

**Out-of-band CI job**, not part of `quality` on PRs. Runs nightly (or weekly — to be discussed) on a `schedule:` trigger in `.github/workflows/`. Reuses the Phase 7 test infra (`test/api-clients/`) but runs the same tests against `global.fetch` instead of mocked fixtures.

### Sketch

```
.github/workflows/contract-check.yml
  on: schedule (cron) + workflow_dispatch
  job: run vitest with VITEST_LIVE_CONTRACT=1
       → test/api-clients/*.test.ts switches from fixture-replay to live fetch
       → asserts parse succeeds, no missing-required-field errors
       → on failure: open or update a sticky issue tagged `contract-drift`
```

### Open questions (for team discussion)

1. **Auth.** Shikimori user-rate/friends/history endpoints require OAuth. Do we burn a dedicated bot account's token (refresh stored in GitHub secret) or restrict the live test to anonymous endpoints only (anime details, search) and accept narrower coverage?
2. **Cadence.** Nightly is the obvious default. Weekly is gentler on upstream and on our notification budget; daily is overkill for APIs that change at quarterly cadence.
3. **Failure UX.** Sticky issue (recreate on each failure) vs. auto-issue per failure vs. Telegram ping via the existing bot (see [[reference_telegram_bot]] / [[project_anime_bot_ghcr]])? Sticky issue is cheapest; Telegram ping is fastest.
4. **Scope.** Just Shikimori + smotret? Add aniskip too once it's promoted past provisional ([[project_aniskip_provisional]])?
5. **What we assert.** Tight (every field that exists today must exist tomorrow) vs. loose (parser doesn't throw, required fields are non-null). Tight catches more; loose has fewer false positives from benign additions.

## Implementation Plan

*High-level only — this issue is a proposal, not a ready-to-implement plan. Detail lands once the team discussion settles the open questions above.*

1. **Test infra:**
   - [ ] Add `VITEST_LIVE_CONTRACT` env-flag branch to `test/api-clients/shikimori.test.ts` + `smotret-api.test.ts` (Phase 7 slice 7b.2 — already creating these files). When set, swap fixture replay for live `fetch`.
2. **CI:**
   - [ ] `.github/workflows/contract-check.yml` — `schedule:` cron + `workflow_dispatch`; sets `VITEST_LIVE_CONTRACT=1`; runs `npm run test -- test/api-clients/`.
   - [ ] Failure handling per the team decision in open question #3.
3. **Secrets (if user-scoped endpoints in scope):**
   - [ ] Provision a dedicated Shikimori OAuth refresh token in GitHub repo secrets; document the refresh procedure.

## Files to Touch

- **New:** `.github/workflows/contract-check.yml`.
- **Modify:** `test/api-clients/shikimori.test.ts`, `test/api-clients/smotret-api.test.ts` (env-flag branch — both files originate in Phase 7 slice 7b.2).
- **Documentation:** Short note in `docs/testing.md` (created in Phase 7) describing the live contract job, when it runs, what to do when it fails, and the secret-refresh procedure.

## Testing Strategy

- Manual: trigger `workflow_dispatch` once on `main` after landing; confirm the live tests pass against current upstream.
- Mutation: temporarily flip a parser assertion to expect a non-existent field and confirm the cron fails noisily.

## Risks & Edge Cases

- **Upstream rate limits / TOS.** Daily-or-less requests to a handful of endpoints is well within polite usage, but worth a one-line check of each API's TOS before enabling.
- **Test bot account compromise.** If a Shikimori bot account's refresh token leaks via the workflow logs, it can mutate that account's rates. Mitigation: bot account owns no real watchlist; rotate token on any incident; consider read-only OAuth scope if Shikimori supports it.
- **Notification fatigue.** A noisy job that pages on every transient 500 trains us to ignore it. Mitigation: 2–3 consecutive failures before alerting; surface upstream-5xx separately from parse failures.
- **False positives on schema additions.** A new optional field is not a regression. Mitigation: assert only on required-field presence + parse success, not on full-shape equality (the "loose" option in open question #5).

## Out of Scope

- Replacing the Phase 7 fixture replay tests — they remain the PR-blocking floor; this is a complementary signal.
- Contract checks against any other external service (e.g. ffbinaries, GitHub releases for auto-update).
- Auto-refreshing the committed fixtures from the live response — too magic; humans should review schema changes before pinning new fixtures.

---

Complements #140 (Phase 7 slice 7b.2). Part of the broader test-coverage effort tracked under #84.

akvasha:
## Summary

Implements the live-API contract check approved in the #141 review. Fixture-replay tests under `test/api-clients/` keep our parsers pinned to a recorded response shape; this PR adds a weekly probe against the *live* `shikimori.one` + `smotret-anime.ru` APIs and pages us when a required field disappears or changes type — silent regressions that would otherwise reach a user before us.

Fixes #141.

## What's in here

| Piece | Path |
|---|---|
| Shikimori contract suite (calendar, animes/:id, animes/:id/franchise) | `test/contract/shikimori.contract.test.ts` |
| Smotret-anime contract suite (search → anime → episodes-batch + episode) | `test/contract/smotret-api.contract.test.ts` |
| Contract-only vitest config (longer timeouts, serial, includes `test/contract/**`) | `vitest.contract.config.ts` |
| Default suite excludes `test/contract/**` (PRs stay offline) | `vitest.config.ts` |
| `npm run test:contract` script | `package.json` |
| Weekly workflow + sticky `contract-drift` issue | `.github/workflows/contract-check.yml` |
| Docs entry | `docs/testing.md` |

The `contract-drift` label was created on the repo (orange, autodescribed) — the workflow uses it to find/update the sticky issue.

## How the review settled the open questions

| # | Question | Choice (per @mikkerlo's review) | Reflected in |
|---|---|---|---|
| 1 | Auth | Anonymous endpoints only | Tests never pass a real token; smotret client constructed with `() => ''`. |
| 2 | Cadence | Weekly | `cron: '0 6 * * 1'` (Mon 06:00 UTC). |
| 3 | Failure UX | Auto-created issue, label `contract-drift` | Workflow finds-or-creates a sticky open issue, appends a comment per failure with the run URL. |
| 4 | Scope | Shikimori + smotret only, no AniSkip | Two suites; AniSkip remains provisional per `project_aniskip_provisional`. |
| 5 | Assertion tightness | **Loose** | Tests assert parse-succeeds + required fields non-null with the right primitive type. Upstream adding new optional fields won't trip us. |

## Design deviation worth flagging

The issue's sketch suggested adding a `VITEST_LIVE_CONTRACT=1` branch *inside* the existing `test/api-clients/{shikimori,smotret-api}.test.ts` files. I went with a separate `test/contract/` folder + a dedicated `vitest.contract.config.ts` instead:

- The fixture-replay tests stay PR-blocking and focused on **canned values** from recorded payloads.
- The live contract tests use **loose-shape assertions** — they're a fundamentally different test, and mixing both modes in one file with an env-flag would have made the fixture suite harder to read.
- Same end-state (PRs stay offline; live check runs only on the scheduled workflow).

## Behavior

- **PR runs:** `npm run test` still passes 477/477, contract folder excluded — no network touched. Coverage thresholds + IPC contract guard unchanged.
- **Weekly job:** runs `npm run test:contract`; on parse failure or missing-required-field, opens a sticky `contract-drift` issue (or comments on the existing one) with the run URL, then re-fails the workflow step so the Actions UI doesn't show a misleading green checkmark.
- **Manual dispatch** is wired (`workflow_dispatch:`) so you can trigger a fresh run any time from the Actions tab.

## Test plan

- `npm run test:contract` — verified against the live APIs locally, **4/4 passed in ~4s**. Will also run via `workflow_dispatch` once this lands on `main`.
- Full quality gate green locally: **typecheck · lint · format:check · check:subscription-contract · test (477) · build**.
- I'll trigger the workflow manually on `main` post-merge and report back; if it goes red I'll address before next Monday's scheduled run.

## What's deliberately out of scope (logged in #141)

- "2–3 consecutive failures before paging" debounce — weekly cadence + loose assertions already kill most false-positive sources; revisit if it gets noisy.
- Separating transient 5xx from parse failures.
- Authenticated-endpoint coverage (`whoami`, `user_rates`, `friends`, `history`) — would require provisioning a bot account; not worth the leak risk for now.

### DIFF
--- a/.github/workflows/contract-check.yml
+++ b/.github/workflows/contract-check.yml
@@ -1,0 +1,81 @@
+name: Live API contract check
+
+# Weekly probe against the real shikimori.one + smotret-anime.ru APIs. The
+# fixture-replay tests under test/api-clients/ catch parser drift given the
+# response shape we recorded; this job catches the response shape changing
+# upstream — silent regressions that would otherwise reach a user before us.
+# (#141; complements the fixture suite under test/api-clients/)
+
+on:
+  # Mondays at 06:00 UTC — gentle cadence per the issue review (APIs move
+  # slowly; nightly trains us to ignore the alert).
+  schedule:
+    - cron: '0 6 * * 1'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  issues: write
+
+jobs:
+  contract:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          cache: npm
+
+      - run: npm ci
+
+      - id: run
+        run: npm run test:contract
+        continue-on-error: true
+
+      - name: Open or update sticky contract-drift issue on failure
+        if: steps.run.outcome == 'failure'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REPO: ${{ github.repository }}
+          RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+        run: |
+          # Find an existing open contract-drift issue (sticky).
+          EXISTING=$(gh issue list --label contract-drift --state open --json number \
+            --jq '.[0].number' || echo "")
+
+          BODY=$(cat <<EOF
+          The weekly live API contract check (\`.github/workflows/contract-check.yml\`)
+          failed on $(date -u +%Y-%m-%dT%H:%M:%SZ).
+
+          Run: $RUN_URL
+
+          Likely cause: a response shape change at \`shikimori.one\` or
+          \`smotret-anime.ru\` made one of our parsers' required fields
+          missing or wrong-typed. Update the affected fixture(s) under
+          \`test/fixtures/\` and the corresponding TypeScript interface(s) in
+          \`src/main/shikimori.ts\` or \`src/main/smotret-api.ts\`, then verify
+          \`npm run test:contract\` is green locally before closing this issue.
+          EOF
+          )
+
+          if [ -n "$EXISTING" ]; then
+            echo "Updating existing contract-drift issue #$EXISTING"
+            echo "$BODY" | gh issue comment "$EXISTING" --body-file -
+          else
+            echo "Creating new contract-drift issue"
+            echo "$BODY" | gh issue create \
+              --title "Live API contract check failed" \
+              --label contract-drift \
+              --body-file -
+          fi
+
+      # Surface the failure on the workflow page itself too. The notify step
+      # above creates an issue; this re-fails the job so the green-checkmark
+      # heuristic in the Actions UI doesn't lie about the contract state.
+      - name: Re-fail on contract-test failure
+        if: steps.run.outcome == 'failure'
+        run: |
+          echo "Contract tests failed — see issue created above." >&2
+          exit 1
--- a/test/contract/smotret-api.contract.test.ts
+++ b/test/contract/smotret-api.contract.test.ts
@@ -1,0 +1,109 @@
+// Live contract test against smotret-anime.ru (#141).
+//
+// Excluded from the default `npm run test` suite — runs weekly via
+// `.github/workflows/contract-check.yml` or locally via `npm run test:contract`.
+//
+// Anonymous (empty-token) requests only — sufficient for searchAnime, getAnime,
+// getEpisode, and the new getEpisodesBatch path. getEmbed requires a token and
+// is intentionally out of scope. Assertions are LOOSE per the review on #141:
+// parser doesn't throw, required fields present.
+
+import { describe, it, expect } from 'vitest'
+import { SmotretApi } from '../../src/main/smotret-api'
+
+const api = new SmotretApi(() => '')
+
+function requireString(obj: Record<string, unknown>, key: string): void {
+  expect(obj[key], `expected ${key} to be a string`).toEqual(expect.any(String))
+}
+
+function requireNumber(obj: Record<string, unknown>, key: string): void {
+  expect(obj[key], `expected ${key} to be a number`).toEqual(expect.any(Number))
+}
+
+describe('smotret-anime live contract', () => {
+  // The chain: search → anime → episodes-batch / episode. Each step uses the
+  // ID from the previous, so we don't pin smotret-anime-specific IDs (those
+  // are less stable than MAL ids).
+  it('search → anime → episodes-batch + episode covers the cold-load path', async () => {
+    // 1) searchAnime — non-empty result with the basic AnimeSearchResult shape.
+    const searchResult = await api.searchAnime('Steins;Gate')
+    expect(searchResult.data, 'searchAnime.data must be an array').toBeInstanceOf(Array)
+    expect(searchResult.data.length, 'search "Steins;Gate" must have results').toBeGreaterThan(0)
+
+    // Prefer a result that has episodes (some entries are 0-episode placeholders).
+    const seed = searchResult.data.find((r) => r.numberOfEpisodes > 0) ?? searchResult.data[0]
+    const s = seed as unknown as Record<string, unknown>
+    requireNumber(s, 'id')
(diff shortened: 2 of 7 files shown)


.github/workflows/contract-check.yml
```diff
@@ -0,0 +1,81 @@
...
+      - name: Open or update sticky contract-drift issue on failure
+        if: steps.run.outcome == 'failure'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REPO: ${{ github.repository }}
+          RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+        run: |
+          # Find an existing open contract-drift issue (sticky).
+          EXISTING=$(gh issue list --label contract-drift --state open --json number \
+            --jq '.[0].number' || echo "")
```
mikkerlo, line 46: If no issues are found, `gh issue list` returns `[]` and `jq '.[0].number'` outputs the literal string `"null"`. Bash evaluates `[ -n "null" ]` as true, causing the script to run `gh issue comment "null"` which will fail.

You can fix this by using `// empty` so `jq` outputs an empty string when the array is empty:
```suggestion
            --jq '.[0].number // empty' || echo "")
```

test/contract/smotret-api.contract.test.ts
```diff
@@ -0,0 +1,109 @@
...
+// Excluded from the default `npm run test` suite — runs weekly via
+// `.github/workflows/contract-check.yml` or locally via `npm run test:contract`.
+//
+// Anonymous (empty-token) requests only — sufficient for searchAnime, getAnime,
+// getEpisode, and the new getEpisodesBatch path. getEmbed requires a token and
+// is intentionally out of scope. Assertions are LOOSE per the review on #141:
+// parser doesn't throw, required fields present.
+
+import { describe, it, expect } from 'vitest'
+import { SmotretApi } from '../../src/main/smotret-api'
```
mikkerlo, line 12: nit: Since `vitest.contract.config.ts` sets up the `@main` alias, we can use it here for consistency with the rest of the codebase:
```suggestion
import { SmotretApi } from '@main/smotret-api'
```

mikkerlo (changes requested):
Great implementation! The separate folder and vitest config are clean ways to isolate these from the standard PR suite.

However, I spotted a bug in the GitHub action: if no existing contract-drift issue is open, `jq` will return the literal string `"null"`, which breaks the `[ -n "$EXISTING" ]` check and causes the workflow to crash when trying to comment on issue `#null`. 

I've left a suggestion to fix this, plus a minor nit about imports.

PR #158, opened 2026-05-25
fix(syncplay): drop misleading 'Auto-reconnect disabled' error in disconnect status (#119)

### LINKED ISSUE #119: Test connection in Settings > Watch Together shows misleading "Auto-reconnect disabled" error
**Priority:** Low | **Effort:** Small

## Motivation

In **Settings > Watch Together**, pressing **Test connection** while the **Auto-reconnect on disconnect** toggle is ON surfaces the error message *"Auto-reconnect disabled"*. This is confusing: the user's saved setting is on, but the test claims it's off.

The root cause: `testSyncplayConnection` (now `WatchTogetherTab.vue:39`, previously `SettingsView.vue:1050` on main) hardcodes `autoReconnect: false` in its `syncplayConnect` call so the one-shot test can disconnect cleanly without retrying. When the test ends (success or failure), `src/main/syncplay.ts:701` emits `"Auto-reconnect disabled"` as the disconnect reason — which then bubbles into the test-status callback as `syncplayTestError.value`.

Behavior is unchanged across the Phase 5 5a refactor (#118); this papercut predates it.

## Approach

In `testSyncplayConnection` in `WatchTogetherTab.vue`:

- Don't display the raw `status.error` if it is the test-suppressed reason. Either:
  - **(a)** Set `syncplayTestError.value` to a fixed string like `'Connection closed'` whenever the disconnect arrives during a test (we already know the test disabled auto-reconnect on purpose), or
  - **(b)** After a successful `'ready'` → `dispose()` → `syncplayDisconnect()` sequence, do not re-enter the failed branch even if a late `'disconnected'` broadcast races past `dispose()` (race-safe guard).

Option (a) is the smaller fix; option (b) also resolves a potential race where a late broadcast slips through if `dispose()` and the disconnect event are not strictly serialized.

Bonus: file `syncplay.ts:697-712` only emits the `"Auto-reconnect disabled"` reason for *unexpected* disconnects; an explicit `syncplayDisconnect()` call may want to set `cfg = null` or pass a `userInitiated: true` flag so the reason is omitted entirely.

## Files to Touch

- **Modify:** `src/renderer/src/components/settings/WatchTogetherTab.vue`
- Optionally **Modify:** `src/main/syncplay.ts` (to distinguish user-initiated disconnects from network drops)

## Testing Strategy

- Toggle **Auto-reconnect on disconnect** ON. Click **Test connection** against `syncplay.pl:8999`. Expect either *"Connected successfully"* (if server reachable) or a network-style error — never *"Auto-reconnect disabled"*.
- Repeat with toggle OFF; saved setting must remain off after the test.
- Verify the actual saved `syncplay.autoReconnect` value isn't disturbed by running a test connection.

## Out of Scope

- Any changes to the underlying Syncplay protocol or reconnect strategy.
- Refactoring `testSyncplayConnection` beyond the messaging fix.

akvasha:
## Summary

Fixes the papercut from #119 where pressing **Settings › Watch Together › Test connection** could surface the misleading error *"Auto-reconnect disabled"* even when the user's saved Auto-reconnect setting was ON.

## Root cause

When the Syncplay socket closes while `autoReconnect` is off — either because the user disabled it in Settings, or because `testSyncplayConnection` forces it off so the one-shot probe can disconnect cleanly — `src/main/syncplay.ts:697` writes `'Auto-reconnect disabled'` into `status.error`. That phrase is *why we're not retrying*, not *why we disconnected*; treating it as the error string confuses any UI that just renders `status.error` verbatim (test button **and** the in-player Syncplay popover).

## Changes

- `src/main/syncplay.ts` — split the give-up branch into two reasons. `status.error` is now the neutral `'Connection closed'` when auto-reconnect is intentionally off, and stays `'Max reconnect attempts reached'` when retries are exhausted (that one is genuinely informative). The `room-event` log line keeps the diagnostic `auto-reconnect disabled` mention so debug traces still tell the full story.
- `test/services/syncplay-disconnect-reason.test.ts` — new regression test. Mocks `net`/`tls`, drives the socket `close` event, asserts both branches. Verified to fail on the pre-fix code.

## Behavior

- Test connection w/ Auto-reconnect ON → never shows *"Auto-reconnect disabled"*. Network failures show their underlying reason (or `Connection closed` if the socket dropped silently). Successful tests still show *"Connected successfully"*.
- Test connection w/ Auto-reconnect OFF → same; saved `syncplay.autoReconnect` value remains untouched (the test still forces `autoReconnect: false` over the wire, but that no longer leaks the misleading string back into the UI).
- In-player Syncplay popover also benefits — a manual disconnect (with auto-reconnect off) no longer shows the same misleading error chip.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint` (no new warnings)
- [x] `npm run format:check`
- [x] `npm run test` — 479 tests pass, including the new 2 cases
- [x] `npm run build`
- [x] Regression test verified to fail on main (`Expected: 'Connection closed' / Received: 'Auto-reconnect disabled'`) and pass with the fix.

Fixes #119

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "anime-dl-app",
-  "version": "4.0.6",
+  "version": "4.0.7",
   "description": "Anime episode downloader",
   "main": "./out/main/index.js",
   "engines": {
--- a/src/main/syncplay.ts
+++ b/src/main/syncplay.ts
@@ -695,19 +695,26 @@ export class SyncplayClient extends EventEmitter {
       return
     }
     if (!cfg.autoReconnect || this.reconnectAttempts >= MAX_RECONNECT_ATTEMPTS) {
-      const reason =
-        this.reconnectAttempts >= MAX_RECONNECT_ATTEMPTS
-          ? 'Max reconnect attempts reached'
-          : 'Auto-reconnect disabled'
+      const maxedOut = this.reconnectAttempts >= MAX_RECONNECT_ATTEMPTS
+      // status.error surfaces in the UI (test-connection button, player
+      // popover). 'Auto-reconnect disabled' is *why we're not retrying*, not
+      // *why we disconnected* — showing it as the error confuses users (#119).
+      // Use a neutral 'Connection closed' for the user-initiated-off case;
+      // the room-event log still carries the retry-state for diagnostics.
+      const errorReason = maxedOut ? 'Max reconnect attempts reached' : 'Connection closed'
+      const eventReason = maxedOut ? 'Max reconnect attempts reached' : 'auto-reconnect disabled'
       this.setStatus({
         state: 'disconnected',
         host: cfg.host,
         port: cfg.port,
         room: cfg.room,
         username: cfg.username,
-        error: reason
+        error: errorReason
       })
-      this.emit('room-event', { level: 'warn', text: `Disconnected from Syncplay room: ${reason}` })
+      this.emit('room-event', {
+        level: 'warn',
+        text: `Disconnected from Syncplay room: ${eventReason}`
+      })
       return
     }
     this.reconnectAttempts += 1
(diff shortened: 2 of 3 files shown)


mikkerlo (approved):
The fix appropriately updates the disconnect status reasoning and includes proper test coverage.

LGTM after manual testing

PR #181, opened 2026-05-30
feat(home): persistently dismiss continue-watching cards (#170)

### LINKED ISSUE #170: Home: dismiss continue-watching cards (new behavior, split from redesign)
**Priority:** Low | **Effort:** Small

> Split out of the Refined Dark redesign per review on #162. The redesign issues are restyle-only; this is a **net-new behavior**, so it gets its own issue (same as the per-episode translation override / auto-download toggle splits). Related: epic #160, restyle #162.

## Motivation
The redesign mockup introduced a per-card **dismiss (×)** on Home's Continue-watching list (hover-revealed) plus an "All caught up" empty state once everything is dismissed. There is **no existing dismiss** in `HomeView.vue` today — no handler and no store state — so this is genuinely new interaction + state, not a restyle.

## Approach & Architecture
Renderer-only. Add a dismiss action to each continue-watching card that removes it from the visible list. Decide persistence:
- **In-memory / session-only** (simplest): a `Set` of dismissed keys in `HomeView.vue` or a Pinia store; cleared on app restart. No IPC.
- **Persistent**: store dismissed keys in `electron-store` via a small IPC channel so the choice survives restarts. (Only if the product wants permanent dismissal — confirm before building, as it adds a settings key + IPC.)

Recommend starting **in-memory** unless persistence is explicitly wanted.

## Implementation Plan
   - [ ] Add dismissed-key state + a `dismiss(entry)` handler in `HomeView.vue` (or a tiny store if it must outlive the view).
   - [ ] **Key the dismissed `Set` on `animeId` alone — NOT the existing episode-scoped composite key** (`${kind}:${animeId}:${episodeInt}`). Decision: dismissal hides the show for the rest of the session regardless of new episodes ("dismiss means dismissed"); keying on the episode-scoped key would resurface it on any mid-session episode advance, which defeats the dismiss intent (per @akvasha's review). In-memory state is lost on restart either way.
   - [ ] Hide dismissed entries from the rendered list; reuse the existing empty-state markup for the "all dismissed" case.
2. **Preload & IPC (only if persistent):**
   - [ ] New `home:dismiss-continue-watching` channel + `electron-store` key, wired per `docs/ipc.md`.

## Files to Touch
- **Modify:** `components/views/HomeView.vue` (+ a store/IPC only if persistence is chosen).

## Testing Strategy
- [ ] Renderer test: dismissing a card removes it from the list; the empty state shows once all are dismissed; **and a dismissed show does NOT resurface when its episode advances within the same session** (asserts the `animeId`-scoped key decision above).
- [ ] If persistent: integration test that the dismissed set survives a reload.

## Risks & Edge Cases
- **Reappear rule (resolved):** keyed on `animeId`, a dismissed show stays hidden for the whole session even if a new episode airs or progress changes; it only returns after an app restart (in-memory state). If product instead wants new episodes to resurface a dismissed show, switch the key to the episode-scoped composite — but that's an explicit change, not a side effect.

## Out of Scope
- The visual restyle of the continue-watching cards themselves (that's #162).

mikkerlo:
## Summary
Adds a hover-revealed **× dismiss** to each Home continue-watching card (closes #170).

- Clicking × hides the show. Dismissals **persist across restarts** via a new `dismissedContinueWatching` electron-store key (using the existing generic `getSetting`/`setSetting` — no new IPC channel).
- Keyed on **`animeId:episodeInt`**, so the *same unwatched episode* stays hidden after a relaunch, but once progress advances the card to a **newer episode** (a key never dismissed) the show **resurfaces**.
- The card is wrapped in `.cw-card-wrap` so the dismiss is a sibling `<button>` (not nested inside the card `<button>` — valid HTML); `@click.stop` keeps it from also opening the anime.
- Once everything is dismissed, the existing **"All caught up"** empty state shows; the "Recently added" grid is unaffected.

| Action | Behavior |
|--------|----------|
| Dismiss a card | Hidden + persisted |
| Re-open the app | Stays dismissed (same episode) |
| Watch one more episode | Re-appears (new episode = new key) |

## ⚠️ Deviation from the issue spec
#170 as written called for **in-memory + `animeId`-keyed** dismissal. Per a product decision during implementation, this ships the more capable behavior instead: **persistent across restarts + resurface-on-new-episode** (`animeId:episodeInt`-keyed). Functionally a superset of the issue's intent.

## Store key
Registered `dismissedContinueWatching` (`string[]`, default `[]`) in the store registry (`keys.ts` + `STORE_DEFAULTS` + the pinning test) and documented it in `docs/settings.md`. `docs/renderer.md` Home note updated.

## Tests
`test/renderer/components/home-view.test.ts` (+4 cases): dismiss removes a card **and persists the key**; empty state when all dismissed; **dismissal survives a simulated restart** (getSetting-seeded key hidden on mount); **resurfaces once the episode advances** (the `animeId:episodeInt`-key contract — fails under an `animeId`-only impl). Store-keys pinning test updated. Full suite **531 passed**; typecheck + `format:check` clean.


akvasha:
## Review — solid and well-tested; two points worth a decision before merge

Verified against the diff: CI is green (quality + 3-platform builds + CodeQL), and the implementation is clean.

**What it does (confirmed in the diff):**
- `dismissKey(e) = ` `` `${e.animeId}:${e.episodeInt}` `` ; `visibleEntries` filters dismissed keys; `dismiss()` appends the key and persists via `setSetting('dismissedContinueWatching', [...])`; `onMounted` seeds the set from `getSetting`.
- Empty-state condition correctly switched from `entries` → `visibleEntries`; grid iterates `visibleEntries`.
- Card restructured into `.cw-card-wrap` > `.cw-card` button **+ sibling** `.cw-dismiss` button with `@click.stop` and an `aria-label` — valid HTML (no nested buttons) and the × doesn't also open the anime. 👍
- Store key registered correctly: `STORE_DEFAULTS` + `PERSISTED_STORE_KEYS` + the pinning test + `docs/settings.md` + `docs/renderer.md`.
- Tests (+4) cover persist-on-dismiss, empty-state-when-all-dismissed, survives-restart, and **resurface-on-new-episode** (key `1:13` never dismissed → reappears) — a real characterization test that would fail under an `animeId`-only impl.

### 1. ⚠️ Deliberate deviation from the issue spec
As the description flags: #170 called for **in-memory + `animeId`-keyed**; this ships **persistent (new electron-store key) + `animeId:episodeInt`-keyed**. It's a functional superset and arguably better UX, but it adds a persisted surface — flagging it so the issue owner explicitly signs off on the expanded scope.

### 2. The dismissed-keys list grows unbounded (non-blocking)
`dismiss()` only ever **appends** to `dismissedContinueWatching` — nothing prunes it. Each time a show resurfaces on a new episode and is dismissed again, a new `animeId:episodeInt` key is added and the old one lingers forever, so the array accumulates dead keys for episodes long since passed. Harmless per-dismiss, but worth a prune step — e.g. on load/refresh, drop keys that don't match any current continue-watching entry.

**Minor:** an unavailable card (no `animeId`, rendered `.cw-card.disabled`) still shows a × that no-ops (`dismiss()` early-returns on `!e.animeId`) — cosmetic only.

mikkerlo:
Thanks for the thorough review, @akvasha. Addressing the three points:

1. **Deliberate deviation from the issue spec (persistent + `animeId:episodeInt`-keyed).** Confirmed intentional — this expanded scope is signed off as a product decision. We explicitly chose "persist + resurface on the next episode" over the in-memory + `animeId`-only behavior #170 originally sketched, because it gives the better UX (a dismissal sticks across restarts for that unwatched episode but the show comes back once progress advances). Keeping it as-is.

2. **Unbounded dismissed-keys list — fixed.** Added `pruneDismissed()` in `HomeView.vue`, called after each continue-watching `refresh()` (which covers initial load): it drops any persisted `animeId:episodeInt` key that no longer matches a current entry and re-persists via `setSetting` only when it actually pruned something. Dead keys for long-passed episodes can no longer accumulate. New test asserts a dead key is dropped and the persisted array shrinks (2 → 1).

3. **No-op × on unavailable cards — fixed.** The `.cw-dismiss` button is now `v-if="e.animeId"`, so disabled `.cw-card.disabled` entries no longer render the no-op ×. Covered by a new test.

Pushed as a follow-up commit (`09a6510d`) on `home-dismiss-cw`; typecheck + the home-view/store tests are green.

akvasha:
Holding approval on this one — not for a code-quality reason (the implementation is clean, tested, and green), but because it **deviates from the issue spec** and that needs to be reconciled before merge.

**#170 specified:** in-memory dismissal, keyed on `animeId`.
**This PR ships:** persistent dismissal (new `dismissedContinueWatching` electron-store key), keyed on `animeId:episodeInt`.

It's a functional superset and arguably better UX, but it's an expanded scope (a new persisted store key + resurface-on-new-episode semantics) that wasn't what the issue asked for. Please reconcile one way or the other before this is approved:

1. **Keep the new behavior** → update #170 to reflect the agreed scope (persistent + episode-keyed) and get a maintainer 👍 on the change, so the issue and PR don't disagree; **or**
2. **Match the spec** → revert to in-memory + `animeId`-keyed as written.

Separately, still worth addressing if option 1 is chosen (non-blocking): `dismissedContinueWatching` only ever grows — prune keys that no longer match a current continue-watching entry on load/refresh so the list doesn't accumulate dead episode keys over time.

akvasha (approved):
## Approving ✅

Re-reviewed the latest head including the follow-up commit `09a6510`. My earlier hold was on the spec deviation from #170 (in-memory + `animeId`-keyed → persistent + `animeId:episodeInt`-keyed); the maintainer has explicitly signed off on the expanded scope as a product decision, which reconciles it. Clearing the hold.

**Verified in the latest diff:**
- `pruneDismissed()` runs after every `refresh()`, drops dead `animeId:episodeInt` keys, and re-persists only when it actually pruned — the unbounded-growth concern is addressed.
- `.cw-dismiss` is now `v-if="e.animeId"`, so the no-op × on unavailable cards is gone.
- `dismiss()` guards `!e.animeId`, persists via `setSetting`, and is error-handled; `visibleEntries` filter + empty-state switch are correct; valid HTML (sibling buttons, `@click.stop`, `aria-label`).
- Store key registered in `STORE_DEFAULTS` + `PERSISTED_STORE_KEYS` + pinning test + `docs/settings.md` + `docs/renderer.md`.
- Tests (+6) cover persist, empty-state, survives-restart, dead-key prune (2→1), hidden-button, and the resurface-on-new-episode characterization test. Quality + 3-platform builds + CodeQL all green.

**Non-blocking nit:** #170's body still describes the original in-memory spec — worth a quick edit so the closed issue matches the shipped behavior, but not a merge blocker.

PR #187, opened 2026-06-03
fix(shikimori): load friend avatars past image hotlink protection

### LINKED ISSUE #186: Friend avatars don't load in Friends views (Shikimori image hotlink protection)

## Motivation
Friend avatars never render in the Friends views — both the **Friends Activity** feed (`FriendsActivityView.vue`) and the per-anime **Friends** panel on the detail page (`FriendsPanel.vue`) show empty/broken `` circles instead of the friend's Shikimori avatar. Anime posters in the same rows load fine, which is the tell.

The avatar URLs we get from Shikimori are correct absolute URLs (e.g. `https://shikimori.one/system/users/x48/11.png`), so the data layer is fine — the images just don't load in the renderer.

## Root Cause
Shikimori hotlink-protects the images it serves from `shikimori.one`. A request whose `Referer` is a foreign origin (in our case the renderer's dev-server origin `http://localhost:…` or the packaged `file://` origin) gets a placeholder/empty response instead of the real image. We never set a `Referer`, so every `shikimori.one` image loaded **directly by the renderer** comes back blank.

Why posters look fine but avatars don't:
- Posters in these rows prefer smotret-anime's CDN (`smotretAnime.posterUrlSmall`), which has no hotlink protection.
- Friend avatars are *always* loaded straight from `shikimori.one` with no fallback, so they always fail.

The Shikimori REST calls themselves are unaffected because they run in the **main** process via Node `fetch` (not the Chromium session), so this is purely about renderer-initiated `` loads.

## Approach & Architecture
Install a `session.defaultSession.webRequest.onBeforeSendHeaders` interceptor in the main process that rewrites the `Referer` header to `https://shikimori.one/` for requests to `shikimori.one` (and its image subdomains). This satisfies the hotlink check for all renderer-loaded Shikimori images at once (friend avatars, the user's own avatar, and any `shikimori.one` poster fallback).

This is a main-process-only change — no IPC, preload, or renderer changes required.

## Implementation Plan
   - [ ] Add `src/main/lib/shikimori-images.ts` with a pure `withShikimoriReferer(headers)` helper, the URL filter, and an `installShikimoriReferer(session)` installer.
   - [ ] Call `installShikimoriReferer(session.defaultSession)` early in `bootstrap()` (`src/main/index.ts`).
2. **Preload & IPC:** none.
3. **Renderer (`src/renderer/`):** none.

## Files to Touch
- **New:** `src/main/lib/shikimori-images.ts`, `test/lib/shikimori-images.test.ts`
- **Documentation:** `docs/shikimori.md` (note the Referer interceptor)

## Testing Strategy
- Unit (`test/lib/shikimori-images.test.ts`): `withShikimoriReferer` sets/overrides `Referer` to the Shikimori origin and preserves other headers; `installShikimoriReferer` registers an `onBeforeSendHeaders` listener with the Shikimori URL filter and the listener callback emits rewritten headers. A case asserting a foreign `Referer` is replaced documents the behavior difference.
- Manual: log in to Shikimori, open the Friends Activity feed and a detail page with friends who rated the anime — avatars now render in both dev (`npm run dev`) and a packaged build.

## Risks & Edge Cases
- Scope the URL filter to `shikimori.one` so no other hosts are touched.
- Main-process API `fetch` calls are not affected (different network stack), so authenticated requests keep their normal headers.

## Out of Scope
- A renderer-side broken-image fallback / initials placeholder (could be tracked separately).
- Caching/proxying Shikimori images locally.

akvasha:
## Summary

Fixes #186 — friend avatars never rendered in the Friends views (both the **Friends Activity** feed and the detail-page **Friends** panel); they showed empty circles while anime posters in the same rows loaded fine.

## Root cause

Shikimori hotlink-protects the images it serves from `shikimori.one`. A request whose `Referer` is a foreign origin — the renderer's dev-server `http://localhost:…` or the packaged `file://` origin — gets a placeholder instead of the real image. We never set a `Referer`, so every `shikimori.one` image loaded **directly by the renderer** came back blank.

Posters looked fine because they prefer smotret-anime's CDN (`smotretAnime.posterUrlSmall`), which has no hotlink protection. Friend avatars are always loaded straight from `shikimori.one` with no fallback, so they always failed.

## Changes

- **New `src/main/lib/shikimori-images.ts`** — a pure `withShikimoriReferer(headers)` helper plus `installShikimoriReferer(session)`, which registers a `webRequest.onBeforeSendHeaders` interceptor that rewrites the `Referer` to `https://shikimori.one/` for `shikimori.one` URLs (scoped via a URL filter; image subdomains included).
- **`src/main/index.ts`** — call `installShikimoriReferer(session.defaultSession)` at the start of `bootstrap()`.
- **`docs/shikimori.md`** — document the interceptor under the Friends Activity Feed section.

The Shikimori REST calls are unaffected: they run in the main process via Node `fetch` (a different network stack) and keep their normal auth headers — only renderer-initiated `` loads are touched.

## Behavior

| | Before | After |
|---|---|---|
| Friend avatars (Activity feed + detail panel) | blank circles | render correctly |
| User's own avatar (Settings → Connectors) | also affected | renders correctly |
| Shikimori REST/auth requests | normal | unchanged |

## Test plan

- **Added `test/lib/shikimori-images.test.ts`** (6 cases): `withShikimoriReferer` sets/overrides the `Referer` to the Shikimori origin (incl. the regression case where a foreign `localhost` Referer is replaced — fails on the old behavior), preserves other headers, and doesn't mutate its input; `installShikimoriReferer` registers the listener with the Shikimori URL filter and the listener emits rewritten headers.
- Full quality gate green locally: `typecheck`, `lint` (no new warnings), `format:check`, `test` (564 passing), `test:coverage` (no threshold regressions — `src/main/lib` seam at 96.96%), `build`.
- Manual: logged in to Shikimori → open the Friends Activity feed and a detail page with friends who rated the anime → avatars render in both `npm run dev` and a packaged build.


mikkerlo (approved):
The session webRequest interceptor is the right architectural choice here, cleanly isolating the Referer rewrite to the renderer's Shikimori image requests without bleeding into Node fetch calls.

LGTM after manual testing

PR #191, opened 2026-06-05
fix(skip-detector): FFmpeg WAV-decode fallback when fpcalc can't decode a file

mikkerlo:
## Motivation
Skip detection failed with `FingerprintError: fpcalc exited 3: ERROR: Error decoding audio frame (Invalid data found when processing input)`.

Reproduced with a standard **AAC 44.1kHz / HEVC** MKV (`Bucchigiri_!`). The file itself is fine — the Linux `fpcalc` decodes it without error. The failure is the **Windows `fpcalc.exe`** build's slim, statically-linked FFmpeg aborting mid-decode on a tolerable audio-frame glitch. Worse, that single-episode failure rejected the **entire** show's analysis, even though pairwise matching only needs 2 good episodes.

## Approach
The app already bundles a full FFmpeg build (used for merging and stream-clip fingerprinting). When fpcalc rejects a local file, decode its audio to a temp mono/11025 WAV with that full FFmpeg, then fingerprint the WAV. fpcalc mixes down to mono/11025 internally, so the resulting fingerprint is **byte-identical** to decoding the source directly (verified on the failing file: same `DURATION=1430`, same hashes).

- `getFingerprint` retries via `fingerprintViaWavDecode` on fpcalc failure (gated on `ffmpegPath`).
- `analyzeShow` Phase 1 now **skips** an episode that still can't be decoded instead of throwing, erroring only when fewer than 2 episodes succeed (with a descriptive message).
- Threaded `ffmpegPath` into `AnalyzeOptions`; stream detection inherits the fallback via the existing `DetectStreamOptions.ffmpegPath`.

## Testing
New `test/services/skip-detector-fingerprint.test.ts` (mocks `child_process.spawn`):
- fpcalc rejects a file → falls back to FFmpeg WAV decode and still fingerprints it.
- one undecodable episode is skipped, the rest of the show still analyzes.
- fewer than 2 decodable episodes → descriptive error.

Full unit suite (92 tests) + typecheck pass. Manually verified the WAV fallback produces an identical fingerprint on the original failing file.


akvasha (approved):
LGTM after manual testing on Windows.

Clean, well-scoped fix: the WAV-decode fallback reuses the existing full-FFmpeg + temp-WAV pattern from `fingerprintStreamClip`, the decode params (`-vn -ac 1 -ar 11025 -c:a pcm_s16le`) match what fpcalc resamples to internally, the fallback is correctly gated on `ffmpegPath` + `!signal.aborted` (no ffmpeg spawned after cancel), and temp dirs are cleaned up in `finally`. Degrading a single undecodable episode to a skip — and only erroring when <2 succeed — is the right call given pairwise matching needs just two good episodes. Test coverage hits all three paths (fallback success, skip-without-ffmpeg, <2 descriptive error).

nit (non-blocking): `docs/skip-detection.md` isn't updated. The Fingerprinting / Phase 4 sections describe the pipeline as fpcalc-decodes-the-file-directly; a sentence on the FFmpeg WAV-decode fallback would keep the doc honest per CLAUDE.md's docs-up-to-date rule.

akvasha (commented):
## Overview
When the platform `fpcalc.exe` (with its slim statically-linked FFmpeg) aborts mid-decode on an otherwise-fine file, fingerprinting failed *and* one bad episode rejected the whole show. This PR adds `fingerprintViaWavDecode` (decode audio to a temp mono/11025 WAV with the bundled full FFmpeg, then fingerprint the WAV), retries via that path on fpcalc failure (gated on `ffmpegPath`), makes `analyzeShow` Phase 1 *skip* an undecodable episode and only error when fewer than 2 succeed, and threads `ffmpegPath` through `AnalyzeOptions`. Diagnosis is sound and the fix is well-scoped.

## Strengths
- **Mirrors the existing pattern.** `fingerprintViaWavDecode` closely follows the established `fingerprintStreamClip` (same FFmpeg args, `runChild`, `mkdtemp` + `finally { fs.rm }` cleanup).
- **Correct gating & abort handling.** `getFfmpegPath() || undefined` correctly disables the fallback when FFmpeg isn't downloaded (it returns `''`). The `opts.signal?.aborted` checks ensure cancellation is never swallowed as a "fingerprint failure."
- **Good error message.** The "`N of M succeeded`" failure aggregates per-episode reasons.
- **Solid, hermetic tests.** Mocking `child_process.spawn` and distinguishing ffmpeg vs fpcalc by command/extension is clean; the three cases cover the core behavior, including asserting FFmpeg is invoked *exactly once*.
- Patch version bump (4.1.13 → 4.1.14) is appropriate and will trigger the CI release.

## Suggestions (all non-blocking)
- **Redundant work when fpcalc itself is misconfigured.** `fingerprintLocalFile` falls back on *any* fpcalc failure, including a missing/bad `fpcalcPath` or spawn error — in which case the fallback decodes the WAV and then calls `fingerprintFile` with the *same* broken `fpcalcPath`, wasting an FFmpeg decode per episode before surfacing essentially the same error. Consider narrowing the retry to decode-shaped failures (exit-code 3 / "Error decoding audio frame") rather than spawn failures.
- **Test coverage gap.** No case where `ffmpegPath` *is* set but the FFmpeg decode *also* fails (truly corrupt file) — i.e. the combined "fallback attempted, still skipped" path. Worth one case asserting such an episode is skipped (and <2 such → throws), since that path is most likely to regress silently.
- **Docs.** `docs/skip-detection.md` documents the fingerprinting subsystem in detail (it even calls out the fpcalc/FFmpeg decode behavior). A sentence on the WAV-decode fallback and the "skip undecodable episode, need ≥2" Phase-1 policy would keep it current.

## Correctness / security / performance
- The byte-identical-fingerprint claim is sound given fpcalc's internal mono/11025 downmix, and was verified on the real failing file.
- Temp dir cleaned in `finally`; aborts kill children via the existing plumbing.
- Fallback only runs on the failing episode (FFmpeg invoked once per failure) — no happy-path impact.
- No new external surface; FFmpeg args are fixed over an already-trusted local path.

## Verdict
Approve — correct, well-tested, idiomatic fix for a real failure. Suggestions above are non-blocking polish.

mikkerlo:
Thanks for the thorough review @akvasha — addressed all three non-blocking suggestions in a follow-up commit (a0da42b):

1. **Narrowed the fallback to decode-shaped failures.** `fingerprintLocalFile` now only retries via the FFmpeg WAV decode when fpcalc actually launched and exited non-zero (`isFpcalcDecodeFailure` → `/^fpcalc exited /`). Spawn failures (`fpcalc spawn failed`), a missing binary (`fpcalc path not set`), and aborts rethrow immediately, so we no longer burn a full FFmpeg decode per episode only to hit the same broken `fpcalcPath`.

2. **Closed the test gap.** Added two cases in `skip-detector-fingerprint.test.ts`: an episode where the FFmpeg fallback *is* attempted but the decode also fails (exit 1) is skipped while the rest of the show still analyzes, and the `<2 succeed → descriptive error` variant of that path. The mock now fails ffmpeg when its `-i` input basename starts with `corrupt`.

3. **Docs.** `docs/skip-detection.md` now describes the WAV-decode fallback (Fingerprinting section) and the skip-undecodable / need-≥2 Phase-1 policy (Detection algorithm section).

Typecheck + the 5 skip-detector tests pass.

PR #194, opened 2026-06-22
feat(shikimori): "For You" recommendations tab (#193)

### LINKED ISSUE #193: Add "Recommendations" tab — personalized "what to watch next" feed from Shikimori taste data
**Priority:** Medium | **Effort:** Large

## Motivation

The app already knows a great deal about each user's taste — their Shikimori rates (1–10 scores), completed vs. dropped history, and an aggregated favorite-genres profile (built in `shikimori:get-profile`). Today none of that is turned back into *discovery*: there's a Calendar for shows you already track and a Friends feed, but nothing that answers **"what should I watch next?"** A dedicated **Recommendations** tab closes that loop — surfacing anime the user is statistically likely to enjoy but hasn't added to their list yet, turning the rich Shikimori data we already cache into a reason to keep coming back.

## Approach & Architecture

Shikimori has **no personalized-recommendation endpoint** — only per-anime content similarity (`GET /api/animes/:id/similar`, public/unauthenticated). So the recommendation engine is **assembled locally in the main process** (chosen approach: *local hybrid scoring*), reusing infrastructure that already exists:

- **Seed set** — the user's highest-rated and favorite-genre titles, read from the cached `shikimoriUserRates` + the favorite-genre aggregation already computed for the profile dashboard.
- **Candidate generation** — for each of the top *N* seed titles, fetch `GET /api/animes/:id/similar` (new `getSimilar(malId)` in `src/main/shikimori.ts`, same `shikiFetch` wrapper + 429 handling as the rest of the client). Union the results into a candidate pool, then **exclude anything already on the user's list** (any status in `shikimoriUserRates`, including planned/dropped) so we only recommend genuinely new shows.
- **Scoring & ranking** — a pure, unit-testable `scoreRecommendations()` function ranks each candidate by a weighted blend of the signals the user picked:
  1. **Scores & favorite genres** — genre overlap with the user's favorite-genre vector, weighted by how highly they rated titles in those genres.
  2. **Watch history / completed** — completed titles reinforce their genres; **dropped titles' genres count negatively** so we don't recommend more of what they bailed on.
  3. **Popularity / community score** — Shikimori community `score` + popularity as a tie-breaker so niche/obscure entries don't dominate equally-relevant picks.
  - Each result carries a `reason` ("Because you liked *X*" / "Matches your top genres: Action, Drama") for UI transparency.
- **Resolution to in-app pages** — MAL IDs are resolved to smotret-anime rows via the existing `lookupByMalIds` (reusing the persistent `malIdMap` cache) so each card deep-links into `AnimeDetailView`; unresolved entries show a "Not on smotret-anime" badge and are non-clickable (same pattern as Calendar/Chronology).
- **Caching** — results are expensive to compute (multiple `/similar` calls), so persist them in a new `shikimoriRecommendations` electron-store key with a cache-first IPC + background-refresh broadcast, mirroring the established `shikimori:get-profile` / `get-friends` pattern. Invalidate (or recompute) when the underlying taste data changes: on `shikimori:rates-refreshed` and `shikimori:rate-updated`, and clear on logout. Detail enrichment (genres needed for scoring) reuses the existing `shikimoriAnimeDetails` prefetch cache; candidates lacking cached details are lazily fetched.

The tab is gated on Shikimori login, exactly like Calendar and Friends.

## UI / UX Considerations

- **New sidebar entry** "Recommendations" (or "For You"), rendered in the logged-in `nav-group` in `Sidebar.vue` alongside Calendar/Friends — only visible when `useShikimoriStore.loggedIn` is true.
- **New `RecommendationsView.vue`** under `components/views/`: a `.poster-grid` of recommendation cards reusing the shared `.acard` primitive. Each card shows poster, title, community score badge, and a short **reason chip** ("Because you liked *Steins;Gate*"). Clicking a resolved card opens `AnimeDetailView`.
- A topbar **refresh** button (passes `force`) to recompute on demand, plus a subtle "Updated N min ago" line — consistent with other Shikimori views.
- **Empty / edge states** via the shared `.empty-state` primitive: not-logged-in prompt, "rate a few shows to get recommendations" when the seed set is too small, and a "couldn't reach Shikimori" offline state.
- Optional grouping by reason ("Because you liked X" sections) is **out of scope** for v1 — a single ranked grid keeps it simple.

## Implementation Plan

   - [ ] Add `getSimilar(malId): Promise<ShikiSimilarAnime[]>` to `src/main/shikimori.ts` (`GET /api/animes/:id/similar` via `shikiFetch`).
   - [ ] New `src/main/recommendations.ts`: `buildRecommendations()` (orchestration — seed selection, candidate generation, list-exclusion, detail enrichment, resolution via `lookupByMalIds`) + a **pure** `scoreRecommendations()` (the weighted genre/history/popularity ranking, no I/O, fully unit-testable).
   - [ ] Persist results in a new `shikimoriRecommendations` electron-store key; cache-first read + background refresh; invalidate on rates-refreshed / rate-updated; clear on logout.
   - *New channels start in `src/shared/ipc/channels.ts` (single source of truth), then wire through `src/main/ipc/shikimori.ipc.ts` + `src/preload/index.ts` + `src/preload/types.d.ts`. See `docs/ipc.md`.*
   - [ ] Add `SHIKIMORI_GET_RECOMMENDATIONS: 'shikimori:get-recommendations'` (request/response, accepts optional `force`) to `CHANNELS`.
   - [ ] Add `SHIKIMORI_RECOMMENDATIONS_REFRESHED: 'shikimori:recommendations-refreshed'` to `EVENT_CHANNELS` (background-refresh broadcast, `EventSubscriber<T>` contract).
   - [ ] Add `RecommendationEntry` + payload/return types to `src/shared/types/shikimori.d.ts`; add the `window.api` method + preload signature.
   - [ ] Extend `useShikimoriStore` with `recommendations` state + the `onShikimoriRecommendationsRefreshed` lifetime-scoped subscription (Phase-4 ownership rule).
   - [ ] New `components/views/RecommendationsView.vue` (poster grid + reason chips + refresh + empty states).
   - [ ] Add the gated sidebar entry in `components/shared/Sidebar.vue` and the route/view wiring in `App.vue`.

## Files to Touch
- **New:** `src/main/recommendations.ts`, `src/renderer/src/components/views/RecommendationsView.vue`, `test/recommendations.test.ts`, `test/renderer/recommendations-view.test.ts` (or store test)
- **Modify:** `src/main/shikimori.ts`, `src/main/ipc/shikimori.ipc.ts`, `src/shared/ipc/channels.ts`, `src/shared/types/shikimori.d.ts`, `src/preload/index.ts`, `src/preload/types.d.ts`, `src/renderer/src/stores/shikimori.ts`, `src/renderer/src/components/shared/Sidebar.vue`, `src/renderer/src/App.vue`
- **Documentation:** `docs/shikimori.md` (new Recommendations section), `docs/ipc.md` (new channel + broadcast), `docs/settings.md` (`shikimoriRecommendations` store key), `docs/renderer.md` (new view + store state)

## Testing Strategy
*Automated tests are required.*
- **Unit (main, `test/recommendations.test.ts`):** `scoreRecommendations()` with fixture rates/details — assert (a) a candidate sharing the user's top genres outranks an off-genre one, (b) **a candidate whose genres match a *dropped* title is penalized** (the behavior-difference case: passes only with the negative-history weighting, fails without it), (c) titles already on the user's list are excluded, (d) community score breaks ties between equally-relevant candidates.
- **API-client fixture replay (`test/`):** `getSimilar()` against a recorded `/api/animes/:id/similar` response.
- **IPC contract guard:** the channel-drift test covers the two new channels automatically.
- **Renderer (`test/renderer/`):** store hydrates `recommendations` from the IPC and updates on the `recommendations-refreshed` broadcast; view renders reason chips and the not-logged-in / too-few-ratings empty states.
- **Manual:** logged-in account with a varied rate history → tab populates with sensible, list-excluded picks, reasons read correctly, refresh recomputes, deep-link opens the right anime; logged-out → tab hidden; offline → graceful empty state. Verify on Windows + Linux.

## Risks & Edge Cases
- **Rate-limit pressure** — one `/similar` call per seed title can fan out; cap seed count, sequence requests with the existing throttle/`shikiFetch` 429 handling, and lean on the persistent cache so the tab doesn't refetch on every mount.
- **Cold start** — a user with few/no rated titles has a thin seed set; show the "rate a few shows" empty state rather than garbage recommendations.
- **Missing genre details** — scoring needs `shikimoriAnimeDetails`; candidates with no cached detail must be lazily fetched or scored with reduced confidence, never crash the ranking.
- **Stale cache after taste changes** — make sure invalidation actually fires on rate changes so newly-rated shows shift the feed.
- **Resolution gaps** — many quality recommendations may not exist on smotret-anime; keep them visible (non-clickable badge) rather than silently dropping, so the list doesn't look empty.

## Out of Scope
- Friends'-scores signal (deliberately excluded from v1 ranking — track separately if wanted later).
- Grouping the grid into "Because you liked X" sections (single ranked grid for v1).
- A Home-screen recommendations carousel (tab only for now).
- A full collaborative-filtering / ML model — this is a heuristic hybrid, not a trained recommender.
- Recommending manga or other media types (anime only).

akvasha:
## Summary

Adds a personalized **"For You"** recommendations tab — a "what to watch next" feed surfacing anime the user is likely to enjoy, based on their Shikimori taste. Closes #193.

Shikimori exposes **no personalized-recommendation endpoint** (only per-anime `GET /api/animes/:id/similar`, content-based), so the feed is assembled locally in the main process and cached cache-first like the existing profile/friends payloads.

## How it works

- **Taste profile** — folds the cached rate list into a per-genre affinity map. Sentiment is **score-driven**: high-scored titles push their genres up, low-scored ones push them down, and a drop on its own is neutral (see the Scoring section for why). Every rated MAL id (any status) is the exclusion set — recommendations never include a title already on the user's list.
- **Candidates** — seeds are the user's top-8 scored `completed`/`watching`/`rewatching` titles; the handler fans out `/similar` per seed (sequential, under the rate-limit cap and reused from cache), merging into a candidate pool with per-candidate seed provenance + consensus. Each seed's franchise is also fetched so sibling seasons/sequels are excluded (see Follow-up fixes).
- **Ranking** — weighted blend of genre affinity (primary) + seed score & consensus ("scores & watch history") + community score (popularity tie-break). Each entry gets a reason chip ("Because you liked «X»").
- **Resolution** — `lookupByMalIds` maps each pick to a smotret-anime row so cards deep-link into `AnimeDetailView`; unresolved picks show a "Not on smotret-anime" badge and are non-clickable.

The ranking logic is a set of **pure, I/O-free functions** in `src/main/recommendations.ts` (unit-testable without mocking `electron`/`fetch`); the `shikimori:get-recommendations` handler does the orchestration.

## Scoring formula

Every candidate gets a `matchScore`; the feed is sorted by it descending, with `communityScore` as the tiebreak. The score is a weighted sum of four terms (coefficients are the `W_*` constants in `recommendations.ts`):

```
matchScore(c) =  W_GENRE       * genreScore(c)
              +  W_SEED        * bestSeedScore(c)
              +  W_CONSENSUS   * seedCount(c)
              +  W_POPULARITY  * communityScore(c)
```

| Coefficient | Value | Term it weights | What it captures |
|---|---|---|---|
| `W_GENRE` | **1.0** | `genreScore` | how well the candidate's genres match your taste |
| `W_SEED` | **0.6** | `bestSeedScore` | the score you gave the favorite it came from |
| `W_CONSENSUS` | **0.8** | `seedCount` | how many of your favorites independently point to it |
| `W_POPULARITY` | **0.25** | `communityScore` | Shikimori community score (0–10), a gentle popularity nudge |

**Term definitions:**

- `genreScore(c) = (Σ_{g ∈ c.genres} genreWeight[g]) / sqrt(|c.genres|)` — sum of per-genre affinity, normalized by `sqrt(genre count)` so a candidate can't climb just by listing many genres (genre-stuffing).
- `genreWeight[g] = Σ_{rated titles t with genre g} sentiment(t)` — built once from your whole rate list, where:

  ```
  sentiment(t) =  t.score - 6        if t.score > 0           (7+ positive, ≤5 negative, 6 neutral)
               = +1                  if completed/watching/rewatching but unscored
               =  0                  otherwise (planned / on_hold / dropped-unscored)
  ```

  Sentiment is **score-driven**: a low score is an honest "I disliked this," so it drags the genre down — but a **drop on its own is neutral**. Bailing on a show is often about its quality or pacing, not its genre, so a dropped title only counts against its genres if you *also* scored it low. (Aggregating across your whole list further self-corrects: one weak entry can't flip a genre you otherwise rate highly.)
- `bestSeedScore(c)` — the highest score you gave any seed that surfaced this candidate (0 if none).
- `seedCount(c)` — number of distinct favorites whose `/similar` list contained it (consensus signal).
- `communityScore(c)` — Shikimori's community score for the title, parsed to a number.

**Worked example.** A candidate that surfaced from two of your favorites (scored 9 and 7), shares genres summing to +5 affinity across 4 genres, and has a community score of 8.0:

```
genreScore   = 5 / sqrt(4) = 2.5
matchScore   = 1.0*2.5  +  0.6*9  +  0.8*2  +  0.25*8.0
             = 2.5      +  5.4    +  1.6    +  2.0
             = 11.5
```

**Candidate-genre enrichment.** `genreScore` needs the candidate's genres, which aren't in the detail cache by default (recommendations are unrated). So after merging, a preliminary rank picks the top `RECS_ENRICH_LIMIT` (25) candidates and fetches `getAnimeDetails` for any lacking genres — throttled (250 ms) and cached into `shikimoriAnimeDetails`, so rebuilds and the AnimeDetailView panel reuse them — then the final rank runs with the genre term live. First build pays the fetches; subsequent rebuilds are cheap.

## Follow-up fixes (2nd commit, from review)

Addressing a report that a completed title (*Mushoku Tensei: Isekai Ittara Honki Dasu*) was still recommended, plus feedback on the scoring:

**Exclusion / freshness:**
- **Re-validate the cached feed on every read.** The cached blob outlives rate changes, so the `get-recommendations` cache-hit branch now re-runs `filterOutRated` against the live `shikimoriUserRates` set and persists the trimmed feed — a title you've since rated/completed can never appear even from a stale cache. This was the direct cause of the report.
- **Invalidate on the background `rates-refreshed` path too**, not just in-app `update-rate`, so completions made on shikimori.one clear the cache.
- **Exclude same-franchise entries.** Each seed's franchise (`getFranchise`) is folded into the exclusion set, so other seasons/sequels/recaps of a show you've watched are dropped (a continuation is Chronology, not discovery).

**Scoring:**
- **Removed the blanket dropped-title genre penalty.** Sentiment is now score-driven — a dropped-but-unscored title is neutral, since a drop is often about a show's quality/pacing, not its genre. It only counts against its genres if you also scored it low.
- **Candidate-genre enrichment** (see the Scoring section) so the genre-affinity term actually fires for unrated candidates.

Regression tests: a completed title is scrubbed from a stale feed on read; a scored-low genre is penalized while a merely-dropped one is not.

## Changes

- **Main:** `recommendations.ts` (pure engine: taste profile, ranking, `filterOutRated`), `getSimilar()` client method, `shikimori:get-recommendations` handler + franchise/read-time exclusion + background refresh in `shikimori.ipc.ts`.
- **IPC:** `SHIKIMORI_GET_RECOMMENDATIONS` + `SHIKIMORI_RECOMMENDATIONS_REFRESHED` channels, `RecommendationEntry` type, preload bridge methods.
- **Store:** new `shikimoriRecommendations` electron-store key — invalidated on rate change (in-app + background refresh) so it rebuilds from fresh taste; cleared on logout.
- **Renderer:** `RecommendationsView.vue` (poster grid + reason chips + refresh + empty states), store `recommendations` state + lifetime broadcast subscription, gated sidebar entry, `App.vue` route, `recommendations` added as a stacked library view.
- **Docs:** `shikimori.md`, `ipc.md`, `settings.md`, `renderer.md`.

## Behavior

- **Logged out** → tab hidden; if opened directly, a "Connect to Shikimori" prompt.
- **Too few ratings** → "rate a few shows" hint (distinct from the connect prompt).
- **Offline / API error** → graceful empty state with the error.
- Refresh button recomputes; opening the tab after rating a show rebuilds the feed.

## Test plan

- `test/recommendations.test.ts` — pure-engine units, including the **behavior-difference case** (Drama/liked > Mecha/dropped-unscored > Ecchi/scored-low — proving a drop alone doesn't penalize a genre, only a low score does), score-driven sentiment, on-list exclusion, community-score tiebreak, reason text, result cap, and the stale-cache `filterOutRated` regression.
- `test/api-clients/shikimori.test.ts` — `getSimilar()` fixture replay (parse + URL/auth).
- `test/renderer/stores/shikimori-store.test.ts` — `refreshRecommendations` + the `recommendations-refreshed` broadcast.
- Full gate green locally: `typecheck` · `lint` · `format:check` · `test` (594 passing) · `test:coverage` · `build` · `check:subscription-contract`.
- Version bumped `4.1.14` → `4.2.0`.


src/main/ipc/shikimori.ipc.ts
```diff
@@ -172,6 +183,212 @@ export function register({
...
+        communityScore: r.communityScore,
+        reason: r.reason
+      }
+    })
+
+    store.set('shikimoriRecommendations', entries)
+    return entries
+  }
+
+  function refreshShikimoriRecommendationsInBackground(): void {
```
mikkerlo, line 381: nit: no in-flight dedupe — every cache-hit read and every Refresh click stacks another full rebuild (≤8 franchise + ≤8 similar + up to 25 enrich fetches with 250 ms spacing). This matches the existing friends/profile refreshers so it's fine to ship as-is, but recommendations is by far the most expensive of the three; a shared `let inFlight: Promise<...> | null` that concurrent callers await would cap the cost. Could be a follow-up for all three refreshers at once.

src/main/ipc/shikimori.ipc.ts
```diff
@@ -238,6 +456,26 @@ export function register({
...
   })
 
+  ipcMain.handle(CHANNELS.SHIKIMORI_GET_RECOMMENDATIONS, async () => {
+    if (!store.get('shikimoriUser')) return null
+    const cached = store.get('shikimoriRecommendations') as RecommendationEntry[]
+    if (cached && cached.length > 0) {
+      // The cached feed outlives rate changes, so re-validate it against the
+      // current list before serving — an entry the user has since rated or
+      // completed must never show even if the cache predates that change.
+      const filtered = filterOutRated(cached, ratedMalIdSet())
```
mikkerlo, line 470: nit: if the read-time filter empties the feed entirely (`filtered.length === 0`), this still returns `[]` — the renderer shows the "rate a few shows" empty state while the background rebuild is in flight. Rare in practice (rate changes also clear the cache outright), but falling through to the synchronous rebuild when `filtered` comes back empty would avoid the misleading flash.

src/main/ipc/shikimori.ipc.ts
akvasha: Done in ccaeef6 — `fetchAndCacheShikimoriRecommendations` now memoizes a single in-flight build promise (cleared in `finally`), so concurrent cache-hit reads / Refresh clicks await the same rebuild instead of stacking fan-outs. Scoped to recommendations only (the expensive one); extending the pattern to the friends/profile refreshers can be the follow-up you suggested. Covered by an integration test that gates `/similar`, fires two concurrent reads through the real handler, and asserts a single fan-out.

src/main/ipc/shikimori.ipc.ts
akvasha: Done in ccaeef6 — when `filterOutRated` empties the cached feed the handler now clears the stale cache and falls through to the synchronous rebuild, so the renderer gets fresh entries instead of a misleading empty-state flash. Regression test (`test/integration/shikimori-recommendations.test.ts`) asserts the handler returns the rebuilt feed (not `[]`) when every cached entry has since been rated.

PR #195, opened 2026-06-27
fix(shikimori): migrate API origin to shikimori.io + fix rate-cache target_id

akvasha:
## Summary

Two users reported Shikimori statuses showing **"Planned"** on already-watched anime and a **`Shikimori API error: 404`** when saving a rate. Both trace to a single root cause: **Shikimori migrated its domain off `shikimori.one`** (it now 301-redirects to `shikimori.io`), and Node/Electron's `fetch` (undici) **strips the `Authorization` header on a cross-origin redirect**.

The breakage was asymmetric and therefore confusing:
- **Public GET reads** (anime details, friends, recommendations) transparently follow the redirect and keep working → the panel looks healthy.
- **Authenticated writes** (`PATCH`/`POST /api/v2/user_rates`) arrive at `.io` with no `Authorization` → Shikimori returns **404** (the Save error).
- **Token refresh** (`POST /oauth/token`) breaks across the redirect, so the live `get-rate` handler throws before returning → the detail panel falls back to its `'planned'` default for every title.

I confirmed the migration directly: `shikimori.one/...` → `301` → `shikimori.io/...`, undici follows it, and the `Authorization` header is dropped on that hop. The user's rate data is intact on `shikimori.io`.

## Changes

- **Centralize the origin.** New `src/shared/shikimori.ts` exports `SHIKIMORI_ORIGIN` (`https://shikimori.io`) and the derived image hotlink filters. The API base (`shikimori.ts`), continue-watching, the IPC layer, image referer/filters, and all renderer deep-links now derive from it — no more scattered domain literals. The next migration is a one-line change.
- **Fix a latent rate-cache bug.** `fetchAndCacheShikimoriRates` built cache entries with `target_id: rate.target_id`, but the `/api/users/:id/anime_rates` endpoint nests the id under `anime.id` and has **no top-level `target_id`** — so all cached entries stored a null target and the store's `rateByMalId()` could never match (the offline status mirror was dead). Extracted a pure `buildRateCacheEntries` helper that falls back to `anime.id`.
- **Defense-in-depth:** `rateByMalId` in the store now also falls back to `shikiAnime.id`, so entries already cached on disk (with null `target_id`) resolve immediately without waiting for a re-fetch.
- Updated `docs/shikimori.md` with the domain-centralization rule and the redirect/auth-stripping rationale.

## Behavior

- Before: watched anime show "Planned"; saving a rate fails with `404`.
- After: live status and rate writes work against `shikimori.io`; cached statuses resolve even from pre-fix caches.

## Test plan

- `test/lib/shikimori-rate-cache.test.ts` (new) — derives `target_id` from `anime.id` when the row lacks one (fails on the old straight-copy behavior).
- `test/renderer/stores/shikimori-store.test.ts` — new case: `rateByMalId` matches a legacy entry whose `target_id` is null via `shikiAnime.id`.
- `test/lib/shikimori-images.test.ts` — Referer/filters now assert `shikimori.io`; added a filter-scope regression.
- `test/api-clients/shikimori.test.ts` — new regression asserting requests target the `shikimori.io` origin.
- Full gate green locally: `typecheck`, `lint` (0 errors), `format:check`, `test` (585 passing), `build`.

### DIFF
--- a/docs/shikimori.md
+++ b/docs/shikimori.md
@@ -1,6 +1,8 @@
 # Shikimori Integration
 
-Syncs anime watch status and episode progress with [Shikimori](https://shikimori.one). Uses MAL IDs as the shared key between smotret-anime and Shikimori (`myAnimeListId` field on anime series).
+Syncs anime watch status and episode progress with [Shikimori](https://shikimori.io). Uses MAL IDs as the shared key between smotret-anime and Shikimori (`myAnimeListId` field on anime series).
+
+> **Domain:** Shikimori has migrated origins over time (`shikimori.org` → `shikimori.one` → `shikimori.io`). The origin lives in **one** place — `SHIKIMORI_ORIGIN` in `src/shared/shikimori.ts` — and everything (the API base in `shikimori.ts`, image hotlink filters, renderer deep-links) derives from it. Don't hardcode a domain literal anywhere else. The reason centralization matters: a stale domain 301-redirects to the new host, and Node/Electron `fetch` (undici) **strips the `Authorization` header on a cross-origin redirect** — so public reads silently keep working while every authenticated write 404s, which is hard to diagnose.
 
 ## OAuth Flow (OOB)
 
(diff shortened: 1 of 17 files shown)


mikkerlo (approved):
Reviewed: domain centralization is complete (no remaining `.one`/`.org` literals outside the migration-history comment), the `target_id` fallback is consistent end-to-end (main-process update handler + renderer upsert now both match the `anime.id`-derived target), and the new tests capture the behavior difference. Clean fix — LGTM.

PR #197, opened 2026-06-27
perf(anime-detail): cache-first, non-blocking anime page open (#196)

### LINKED ISSUE #196: Cache-first, non-blocking anime detail page (instant open, background refresh without UI lag)

## Motivation

Opening an anime detail page feels slow and can lag the whole app, even for titles already in the library with everything cached locally. Investigation of the current load path (`AnimeDetailView.vue:190-280`) found the page is **network-first and largely sequential**, so the actual content (the episode list) cannot paint from cache:

- **Episode translations are network-first.** `getEpisodesBatch` (`src/main/ipc/anime.ipc.ts:82-104`) hits smotret-anime and only falls back to the local cache *on error*. The episode rows — the main content — therefore block on a smotret round-trip on every open, even when the data is already cached. smotret's API is the slow link and it sits on the critical path.
- **`getAnime` is awaited inline after the cache render** (`AnimeDetailView.vue:233`). The "refresh" isn't a background update — it blocks the rest of mount (watch-progress, Shikimori, skip-detection all wait behind it) and re-runs the pipeline.
- **`checkFileStatus` runs twice per open** (`AnimeDetailView.vue:225` and `:242`) — two file scans + two `downloadedEpisodesGet` IPC calls for a single open.
- **The first per-anime file scan is synchronous** `readdirSync` on the **main process** (`scanEpisodeFiles`, `src/main/index.ts:302`, via `checkEpisodeFiles:468`). The main process is single-threaded and serializes all IPC for every window, so the first scan stalls the whole app — worse with cold-storage dirs or large folders. The *background* rescan already uses async `fsPromises.readdir` (`src/main/index.ts:397`); the first-hit path does not.
- **Non-library anime have no cache at all** — for *both* the detail and the episodes. `setAnimeCache` is library/downloaded-gated (`src/main/ipc/cache.ipc.ts:20-22`), and `animeCacheService.updateEpisode` is gated by the same `isCachable` check (`src/main/services/anime-cache/index.ts:109`). So `getAnimeCache` returns `null`, episode translations aren't cached either, and the user waits on the full `getAnime` → `getEpisodesBatch` double-network sequence behind a spinner. Cache-first reads can't help these titles; only cold-open parallelization (below) does.

The user is fine with caching and background refreshes happening — they just want the page to open instantly from cache and the refresh to not jank the UI.

## Approach & Architecture

Make the anime page **cache-first with non-blocking background refresh**, mirroring the pattern already used for the Shikimori rate cache (cache-first read + background API refresh + surgical patch).

1. **Cache-first episode translations.** Add a **dedicated cache-first read IPC channel** (e.g. `get-episodes-batch-cached`) that returns whatever translations are already in `animeCacheService` synchronously, with no network. The renderer renders those rows immediately, then kicks off the existing network `getEpisodesBatch` in the background and patches the rows when it returns. Keeping it on a separate channel avoids changing the blast radius/semantics of the existing network-first `GET_EPISODES_BATCH`. **Share the read logic:** the new handler and the existing error-fallback in `GET_EPISODES_BATCH` (`src/main/ipc/anime.ipc.ts:94-101`) must both call one `readCachedEpisodes(animeId, episodeIds)` helper so the two cache reads can't drift.
2. **Background `getAnime` refresh.** When a cached `AnimeDetail` was rendered, don't `await getAnime` inline — fire it as a background refresh that patches `anime.value` (guarded by the existing `loadGeneration`/`disposed` checks), instead of blocking mount and re-running the whole pipeline. Preserve the existing `renderedFromCache` guard (`AnimeDetailView.vue:248`) so a *failing* background refresh stays on `dataSource: 'cache'` and never flips the offline chip after a successful cache paint.
3. **Parallelize the cold open.** When there's no cache (e.g. a non-library anime, which has neither detail nor episodes cached), run `getAnime` and the episode batch concurrently rather than sequentially so the user waits on one round-trip, not two stacked.
4. **Call `checkFileStatus` once per open** instead of from both the cache path and the API path.
5. **Make the first file scan async — narrowly.** Do *not* change `scanEpisodeFiles`'s signature: it's injected synchronously and called by skip-analysis (`src/main/services/skip-analysis/index.ts:194`, `:266`) and cold-storage (`src/main/services/cold-storage/index.ts:722`), so flipping it to `Promise` ripples async through unrelated services. Instead, **extract a shared `scanEpisodeFilesAsync`** from the async body that already exists inside `backgroundRescan` (`src/main/index.ts:383-…`, via `fsPromises.readdir`), and route only `checkEpisodeFiles`'s first-hit branch (`src/main/index.ts:468`) and `backgroundRescan` through it. The synchronous `scanEpisodeFiles` stays for skip-analysis/cold-storage.
6. **Dedupe concurrent file scans.** Today `checkEpisodeFiles` scans-then-sets `fileCheckCache` synchronously on a miss (`src/main/index.ts:467-469`), so concurrent misses can't overlap. Once the first-hit path awaits, two opens of the same `animeName` (back-nav or any concurrent caller) both miss → both scan → both write (redundant disk scans, last-writer-wins). Item 4 only kills the in-component double, not cross-open races. Add an in-flight `Map<string, Promise<FileCheckResult>>` keyed by `animeName` so concurrent misses share one scan.

## UI / UX Considerations

No new UI surfaces. Behavioral change only:
- The episode list paints immediately from cache on open for previously-viewed anime, then quietly updates if the network refresh changes anything.
- The existing `dataSource` / "Working offline" affordance and loading states must stay coherent — a cache-first render should not flash the offline chip when the background refresh is about to succeed.

## Implementation Plan

   - [ ] Add a `readCachedEpisodes(animeId, episodeIds)` helper that maps ids → `animeCacheService.getEntry(animeId).episodes[id]` (filtering undefined); refactor the existing `GET_EPISODES_BATCH` error-fallback (`anime.ipc.ts:94-101`) to call it, and back the new cache-first channel with it.
   - [ ] Extract `scanEpisodeFilesAsync` from `backgroundRescan` (`src/main/index.ts:383-…`); route `checkEpisodeFiles`'s first-hit branch (`:468`) + `backgroundRescan` through it. Leave the sync `scanEpisodeFiles` untouched for skip-analysis/cold-storage.
   - [ ] Add an in-flight `Map<string, Promise<FileCheckResult>>` in `checkEpisodeFiles` so concurrent misses for the same `animeName` share one scan and one cache write.
   - *New IPC channel starts in `src/shared/ipc/channels.ts` (single source of truth), then wires through `src/main/ipc/anime.ipc.ts` + `src/preload/index.ts` + `src/preload/types.d.ts`. See `docs/ipc.md`.*
   - [ ] Add `GET_EPISODES_BATCH_CACHED` (name TBD) to `CHANNELS`, the handler in `anime.ipc.ts` (backed by `readCachedEpisodes`), the `window.api.*` method in `src/preload/index.ts`, and its signature in `src/preload/types.d.ts`.
   - [ ] `FILE_CHECK_EPISODES` handler signature/return stays the same (`src/main/ipc/file.ipc.ts:13`) — the async change is internal to `checkEpisodeFiles`, which can stay sync-returning by serving the cache hit immediately and only awaiting on a miss; decide whether `checkEpisodeFiles` returns a `Promise` on miss or keeps a sync cache-hit fast path.
   - [ ] Rework `AnimeDetailView.vue` `onMounted` (`:190-280`): cache-first render of anime + episodes, single `checkFileStatus`, background (non-awaited) `getAnime` refresh, parallel cold-open path.
   - [ ] Update `use-episode-list.ts` `loadPageEpisodes` (`src/renderer/src/composables/use-episode-list.ts:276-318`) to support a cache-first read then background network patch without double-fetching.

## Files to Touch
- **New:** none required beyond the new channel wiring (no new module strictly needed; implementer may add a small helper).
- **Modify:** `src/renderer/src/components/views/AnimeDetailView.vue`, `src/renderer/src/composables/use-episode-list.ts`, `src/renderer/src/composables/use-episode-downloads.ts` (checkFileStatus call sites), `src/main/ipc/anime.ipc.ts`, `src/main/index.ts` (`scanEpisodeFiles`/`checkEpisodeFiles`), `src/shared/ipc/channels.ts`, `src/preload/index.ts`, `src/preload/types.d.ts`.
- **Documentation:** `docs/ipc.md` (new channel), `docs/data-flow.md` (episode load path), and `docs/architecture.md`/`docs/storage.md` if the file-scan path changes.

## Testing Strategy
- **Cache-first episodes (renderer composable test):** `loadPageEpisodes` renders cached translations without awaiting a network `getEpisodesBatch`, then patches when the background fetch resolves. Add a case that fails on the current network-first behavior (asserts the network mock is *not* awaited before rows are present).
- **Single `checkFileStatus` (renderer test):** opening the detail view invokes `fileCheckEpisodes` / `downloadedEpisodesGet` once, not twice.
- **Background `getAnime` (renderer test):** a cache-rendered open does not block on `getAnime`; mount-tail work (watch-progress/Shikimori) runs before `getAnime` resolves.
- **Async file scan (main/lib test):** first-hit `checkEpisodeFiles` resolves the same `FileCheckResult` as the old sync path (characterization), now via the shared `scanEpisodeFilesAsync`; assert the sync `scanEpisodeFiles` consumed by skip-analysis/cold-storage is unchanged.
- **Concurrent-scan dedupe (main/lib test):** two concurrent `checkEpisodeFiles` calls for the same `animeName` on a cache miss trigger exactly one underlying scan (fails without the in-flight `Map`).
- **Shared cache read (main test):** `GET_EPISODES_BATCH` error-fallback and the new cache-first channel return identical results for the same cached input (guards against drift).
- **Manual:** open a cached library anime (instant episode list), a cold non-library anime (parallel fetch, no double wait), and an offline/cache-only case (no offline-chip flicker). Verify on Windows + Linux; test with a large download dir / cold-storage path to confirm no main-process stall.

## Risks & Edge Cases
- **Stale-then-fresh flicker:** cache-first rows must reconcile cleanly with the background refresh (translation list changes, episode added/removed) without losing the user's in-session translation override or scroll/focus position (`applyFocusEpisode`).
- **`loadGeneration`/`disposed` races:** background `getAnime` and background episode refresh must respect the existing generation guards so a fast back-navigation doesn't patch a stale view.
- **Async file scan ordering:** making the first scan async lets two opens of the same anime race on `fileCheckCache` — addressed by the in-flight-promise dedupe (item 6). The shared-extract approach (item 5) deliberately leaves the sync `scanEpisodeFiles` in place for skip-analysis/cold-storage so their synchronous assumptions don't break; verify no caller accidentally switches to the async variant.
- **`dataSource`/offline chip correctness** when rendering from cache before the network result is known.
- Quality probing (`probeSelectedQualities`) and the `backgroundQualityProbe` full-scan should remain off the critical path and not regress.

## Out of Scope
- Quality-probe performance rework (`probeSelectedQualities` / full-scan throttling) beyond keeping it non-blocking.
- Changing smotret-anime API endpoints or caching TTLs (`ANIME_DETAIL_CACHE_TTL_MS`).
- Caching anime details for *all* non-library anime persistently (track separately if desired; this issue only needs the cold open to parallelize).
- Player / download-pipeline performance.

mikkerlo:
Closes #196.

## Summary

Opening an anime detail page was network-first and largely sequential, so the episode list couldn't paint from cache even for previously-viewed titles, and the first per-anime file scan stalled the single-threaded main process. This makes the page **cache-first with non-blocking background refresh**.

## Changes

- **Cache-first episode translations.** New `get-episodes-batch-cached` IPC channel returns whatever translations `animeCacheService` already holds (no network). The renderer paints those rows immediately, then background-refreshes via the existing network `get-episodes-batch` and patches. A shared `readCachedEpisodes(animeId, episodeIds)` helper backs both the new channel and the existing network-first error-fallback so the two cache reads can't drift.
- **Background `getAnime` refresh.** After a cache paint, `getAnime` runs as a non-blocking background refresh (`refreshAnimeInBackground`) instead of being awaited inline — the mount tail (watch-progress / Shikimori / skip-detection) no longer waits on it. The `renderedFromCache` guard is preserved: a *failing* refresh keeps `dataSource: 'cache'` and never flips the offline chip after a good paint.
- **Cold open parallelized.** With no cache (e.g. a non-library anime), the detail is fetched and then the episode batch + file scan run concurrently (`Promise.all`). The episode batch can't precede the detail — episode IDs come from the detail response — so the parallelism is batch ∥ file-scan, not batch ∥ detail.
- **Single `checkFileStatus` per open** (was running twice).
- **Async, narrowly-scoped file scan.** Extracted the scan cache into `src/main/lib/episode-file-scan.ts` (`createEpisodeFileScanner`). The first-hit path + background rescan now use an async `scanEpisodeFilesAsync` (`fsPromises.readdir`) so a cold scan never blocks the main event loop. The **synchronous** `scanEpisodeFiles` is left untouched for skip-analysis / cold-storage, which call it synchronously — no async ripple into those services.
- **Concurrent-scan dedupe.** An in-flight `Map<string, Promise<FileCheckResult>>` keyed by `animeName` means concurrent misses (back-nav, parallel openers) share one disk scan + one cache write.

Non-library anime stay cache-less for episodes too (`updateEpisode` is `isCachable`-gated, `anime-cache/index.ts:109`) — the cold-open parallelization is their fix. The `isCachable` gate is deliberately **not** loosened (out of scope, per the issue thread).

## Test plan

- `test/lib/episode-file-scan.test.ts` — async-scan characterization (== sync result), first-hit filtering, **concurrent-scan dedupe (one scan)**, background-rescan broadcast-on-change, invalidate/clear.
- `test/services/anime-ipc-cached-episodes.test.ts` — shared cache-read parity (cached channel vs network-first fallback identical), missing-id filtering.
- `test/renderer/components/anime-detail-load.test.ts` — single `checkFileStatus` per open, cache-first paint, backgrounded `getAnime` (mount-tail unblocked; offline chip persists on failure, clears on success).
- `test/renderer/composables/use-episode-list.test.ts` — cache-first paint + no double-fetch.
- Full gate green locally on the rebased branch: `typecheck`, `lint` (0 errors), `format:check`, `test` (600 passing), `test:e2e` (4 passing).

### DIFF
--- a/src/main/lib/episode-file-scan.ts
+++ b/src/main/lib/episode-file-scan.ts
@@ -1,0 +1,251 @@
+// Episode-file scan cache backing `CHANNELS.FILE_CHECK_EPISODES` (#196).
+//
+// Extracted out of `src/main/index.ts` so the cache-first / async-first-scan /
+// concurrent-dedupe behavior is unit-testable (index.ts is excluded from the
+// coverage gate and pulls in electron). The cache + helpers live here because
+// `coldStorageService` and `skipAnalysisService` also feed off the synchronous
+// `scanEpisodeFiles` scanner.
+//
+// Two scanners, deliberately:
+//   - `scanEpisodeFiles`      — synchronous (`readdirSync`), injected into
+//     skip-analysis + cold-storage which call it synchronously.
+//   - `scanEpisodeFilesAsync` — `fsPromises.readdir`, used by the
+//     `checkEpisodeFiles` first-hit path + the background rescan so the first
+//     per-anime scan never stalls the single-threaded main process.
+// Both fold filenames through the shared `accumulateEpisodeFiles` parser so the
+// author-tag / legacy-filename matching can't drift between them.
+
+import * as fs from 'fs'
+import * as fsPromises from 'fs/promises'
+import * as path from 'path'
+
+/**
+ * Map of episode base-name (sanitized `${anime} - NN`) → matched files on disk.
+ * Produced by the episode-file scanner and (after `filterScanResult`) returned
+ * verbatim by `CHANNELS.FILE_CHECK_EPISODES`.
+ */
+export type FileCheckResult = Record<
+  string,
+  { type: 'mkv' | 'mp4'; filePath: string; translationId?: number; author?: string }[]
+>
+
+/**
+ * Fold one anime directory's filenames into the running scan result. Matches
+ * author-tagged `Name - NN [Author].mkv` first, then legacy un-tagged
+ * `Name - NN.mkv`. Shared by the sync + async scanners.
+ */
+export function accumulateEpisodeFiles(
+  result: FileCheckResult,
+  animeDir: string,
+  files: string[]
+): void {
+  for (const file of files) {
+    const match = file.match(/^(.+?) \[(.+?)\]\.(mkv|mp4)$/)
+    if (match) {
+      const base = match[1]
+      const author = match[2]
+      const ext = match[3] as 'mkv' | 'mp4'
+      if (!result[base]) result[base] = []
+      const existing = result[base].find((e) => e.author === author)
+      if (existing) {
+        existing.type = ext
+        existing.filePath = path.join(animeDir, file)
+      } else {
+        result[base].push({ type: ext, filePath: path.join(animeDir, file), author })
+      }
+      continue
+    }
+
+    const legacyMatch = file.match(/^(.+)\.(mkv|mp4)$/)
+    if (legacyMatch) {
+      const base = legacyMatch[1]
+      const ext = legacyMatch[2] as 'mkv' | 'mp4'
+      if (!result[base]) result[base] = []
+      const hasAuthorVersion = result[base].some((e) => e.author)
+      if (!hasAuthorVersion || !result[base].some((e) => !e.author)) {
+        const existing = result[base].find((e) => !e.author)
+        if (existing) {
+          existing.type = ext
+          existing.filePath = path.join(animeDir, file)
+        } else {
+          result[base].push({ type: ext, filePath: path.join(animeDir, file) })
+        }
+      }
+    }
+  }
+}
+
+/** Narrow a full per-anime scan down to the requested episode-ints. */
+export function filterScanResult(
+  fullResult: FileCheckResult,
+  animeName: string,
+  episodeInts: string[],
+  sanitizeFilename: (name: string) => string
+): FileCheckResult {
+  const baseMap = new Map<string, string>()
+  for (const epInt of episodeInts) {
+    const padded = epInt.padStart(2, '0')
+    const base = sanitizeFilename(`${animeName} - ${padded}`)
+    baseMap.set(base, epInt)
+  }
+
+  const filtered: FileCheckResult = {}
+  for (const [base, files] of Object.entries(fullResult)) {
+    const epInt = baseMap.get(base)
+    if (epInt) filtered[epInt] = files
+  }
+  return filtered
+}
+
+export interface EpisodeFileScanDeps {
+  /** Directories to scan for an anime (download dir + optional cold-storage dir). */
+  getDirsToScan: () => string[]
+  sanitizeFilename: (name: string) => string
+  /** Emitted by a background rescan when the on-disk set changed since the cache. */
+  onEpisodesChanged: (animeName: string, result: FileCheckResult) => void
+}
+
+export interface EpisodeFileScanner {
+  /** Synchronous full scan — consumed by skip-analysis + cold-storage. */
+  scanEpisodeFiles: (animeName: string) => FileCheckResult
+  /** Async full scan — backs the first-hit path + background rescan. */
+  scanEpisodeFilesAsync: (animeName: string) => Promise<FileCheckResult>
+  /**
+   * Cache-first episode-file check. Cache hit → serve immediately + kick a
+   * background rescan; cache miss → async scan, deduped so concurrent misses
+   * for the same anime share one scan + one cache write.
+   */
+  checkEpisodeFiles: (animeName: string, episodeInts: string[]) => Promise<FileCheckResult>
+  backgroundRescan: (animeName: string) => Promise<void>
+  /** Drop a single anime's cache entry (after a download/merge/delete). */
+  invalidate: (animeName: string) => void
+  /** Drop the entry whose key sanitizes to `dirName` (after a path-keyed delete). */
+  invalidateByDirName: (dirName: string) => void
+  /** Clear the whole cache (before a bulk move-to-cold). */
+  clear: () => void
+}
+
+export function createEpisodeFileScanner(deps: EpisodeFileScanDeps): EpisodeFileScanner {
+  const { getDirsToScan, sanitizeFilename, onEpisodesChanged } = deps
+  const fileCheckCache = new Map<string, FileCheckResult>()
+  // In-flight scans keyed by animeName: concurrent misses (back-nav, parallel
+  // openers) share one disk scan + one cache write instead of double-scanning.
+  const inFlight = new Map<string, Promise<FileCheckResult>>()
+
+  function scanEpisodeFiles(animeName: string): FileCheckResult {
+    const animeDirName = sanitizeFilename(animeName)
+    const result: FileCheckResult = {}
+    for (const dir of getDirsToScan()) {
+      const animeDir = path.join(dir, animeDirName)
+      if (!fs.existsSync(animeDir)) continue
+      let files: string[]
+      try {
+        files = fs.readdirSync(animeDir)
+      } catch {
+        continue
+      }
+      accumulateEpisodeFiles(result, animeDir, files)
+    }
+    return result
+  }
+
+  async function scanEpisodeFilesAsync(animeName: string): Promise<FileCheckResult> {
+    const animeDirName = sanitizeFilename(animeName)
+    const result: FileCheckResult = {}
+    for (const dir of getDirsToScan()) {
+      const animeDir = path.join(dir, animeDirName)
+      let files: string[]
+      try {
+        files = await fsPromises.readdir(animeDir)
+      } catch {
+        continue
+      }
+      accumulateEpisodeFiles(result, animeDir, files)
+    }
+    return result
+  }
+
+  function normalize(r: FileCheckResult): FileCheckResult {
+    const sorted: FileCheckResult = {}
+    for (const key of Object.keys(r).sort()) {
+      sorted[key] = [...r[key]].sort((a, b) => (a.author || '').localeCompare(b.author || ''))
+    }
+    return sorted
+  }
+
+  async function backgroundRescan(animeName: string): Promise<void> {
+    const result = await scanEpisodeFilesAsync(animeName)
+    const cached = fileCheckCache.get(animeName)
+    const newJson = JSON.stringify(normalize(result))
+    const cachedJson = cached ? JSON.stringify(normalize(cached)) : ''
+    if (newJson !== cachedJson) {
+      fileCheckCache.set(animeName, result)
+      onEpisodesChanged(animeName, result)
+    }
+  }
+
+  async function checkEpisodeFiles(
+    animeName: string,
+    episodeInts: string[]
+  ): Promise<FileCheckResult> {
+    const cached = fileCheckCache.get(animeName)
+    if (cached) {
+      backgroundRescan(animeName).catch((err) => console.error('Background rescan failed:', err))
+      return filterScanResult(cached, animeName, episodeInts, sanitizeFilename)
+    }
+
+    let scan = inFlight.get(animeName)
+    if (!scan) {
+      // `scan` is assigned synchronously before either callback can run, so it's
+      // safe to compare against the live in-flight entry inside them. If an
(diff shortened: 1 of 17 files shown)


src/main/lib/episode-file-scan.ts
```diff
@@ -0,0 +1,239 @@
...
+    if (cached) {
+      backgroundRescan(animeName).catch((err) => console.error('Background rescan failed:', err))
+      return filterScanResult(cached, animeName, episodeInts, sanitizeFilename)
+    }
+
+    let scan = inFlight.get(animeName)
+    if (!scan) {
+      scan = scanEpisodeFilesAsync(animeName)
+        .then((result) => {
+          fileCheckCache.set(animeName, result)
```
akvasha: `invalidate()` can be clobbered by an in-flight scan, briefly serving stale file status.

This `.then` re-`set`s the cache unconditionally, with no check that the entry was invalidated while the scan was running. `invalidate()` (download-complete, merge-complete, cold-storage move, cleanup-delete, `file:open`-missing) only does `fileCheckCache.delete(animeName)` and never touches `inFlight`.

Scenario: the user is viewing anime X while a `checkEpisodeFiles` miss-scan is in-flight, and a download/merge for X completes → `invalidate(X)` deletes the entry, then this `.then` re-`set`s the **pre-merge** result. The next open serves that stale entry (missing the just-downloaded episode's ✓). `backgroundRescan` detects the diff and broadcasts a correction on the following check, so it self-heals within one cycle — but the old synchronous code had no such window, so this is a (minor) regression.

Cheap fix: have `invalidate`/`clear` also `inFlight.delete(animeName)`, or tag each scan with a generation token and skip the `.then` `set` if the key was invalidated during the scan.

src/renderer/src/composables/use-episode-list.ts
```diff
@@ -312,7 +310,63 @@ export function useEpisodeList(deps: {
...
+    } catch {
+      // No cache-first channel / read failed — fall through to the network.
+    }
+
+    if (paintedFromCache) {
+      applyAuthorDefault()
+      loadingEpisodes.value = false
+      void probeSelectedQualities()
+      // Background-refresh the rows we painted from cache; patch on return.
+      void networkFetchEpisodes(toLoad, ids)
```
akvasha, line 371: Partial episode cache triggers a duplicate network fetch.

When the anime *detail* is cached but only a **subset** of episode translations are cached, this fires `networkFetchEpisodes(toLoad, ids)` for the full `toLoad`. Then `refreshAnimeInBackground` calls `loadPageEpisodes` again; its `toLoad` filter still includes the un-cached episodes (they weren't painted, so `!episodes.has(id)`), `getEpisodesBatchCached` returns empty for them, and it `await`s a second `getEpisodesBatch` covering that same subset → those ids get fetched twice.

The common paths (full episode cache, or no cache at all) are unaffected. Low severity — wasted network/CPU on a partial-cache open, not incorrect output. Could be avoided by synthesizing/marking the un-cached ids after the cache paint, or skipping the redundant `loadPageEpisodes` in the background refresh when nothing in the detail changed.

mikkerlo:
Thanks for the review @akvasha — both addressed in `98ac1aa`.

**1. `invalidate()` clobbered by an in-flight scan (`episode-file-scan.ts`).** The miss-path scan now writes the cache *only if it's still the current in-flight entry*: the `.then` guards with `if (inFlight.get(animeName) === scan) fileCheckCache.set(...)`, and the `.finally` cleanup uses the same `=== scan` check so a superseded scan can't evict a newer in-flight entry. `invalidate` / `invalidateByDirName` / `clear` now also drop the matching `inFlight` entry. So an `invalidate` (download/merge/move/delete) landing mid-scan no longer gets the stale pre-change result written back over the delete — the awaiting caller still gets the fresh result returned, only the stale *write* is suppressed. Regression test added (invalidate while a miss-scan is pending → next check runs a fresh scan; fails on the old unconditional-`set`).

**2. Partial-cache duplicate fetch (`use-episode-list.ts`).** Added a composable-scoped `inFlightEpisodeIds` set — `networkFetchEpisodes` adds ids before the await and removes them in a `finally`, and the `toLoad` filter now excludes in-flight ids (`!episodes.has(id) && !inFlightEpisodeIds.has(id)`). The post-`getAnime` second `loadPageEpisodes` sees the un-cached subset already in flight and drops it, so those ids are fetched once. Display/paint behavior is unchanged (no stub synthesis), generation guard preserved. Regression test added (partial cache + two overlapping loads → each un-cached id fetched at most once; fails on the old code).

Full gate green: typecheck, lint (0 errors), format:check, **602 tests**, coverage thresholds hold. Re-requesting review.

akvasha (approved):
Re-reviewed the delta since `07fc35c` — both review comments are addressed correctly, each with a dedicated regression test.

- **File-scan stale-write race** (`episode-file-scan.ts`): `invalidate`/`invalidateByDirName`/`clear` now also drop the `inFlight` entry, and the in-flight `.then`/`.finally` are guarded by an identity check so a superseded scan can neither repopulate the cache after an invalidate nor evict a newer scan. Traced clean; covered by *"invalidate during an in-flight miss-scan suppresses the stale cache write."*
- **Partial-cache duplicate fetch** (`use-episode-list.ts`): the new `inFlightEpisodeIds` set (added before fetch, removed in `finally`) plus the `toLoad` exclusion stops the post-`getAnime` background refresh from re-fetching the subset already in flight, while still retrying on failure. Covered by *"does not double-fetch un-cached episodes across overlapping loads (partial cache)."*

No new issues introduced. Resolving both threads. LGTM 👍

PR #199, opened 2026-06-27
fix(player): correct subtitle clock after MKV seek (Option 1, renderer self-correction)

### LINKED ISSUE #198: Fix subtitle/audio desync after seeking in cached MKV (MSE timestampOffset re-mapping wrong when keyframeTime > 0)

## Motivation

When the built-in MSE player seeks (skips) within a **cached MKV**, the audio/video and the ASS subtitles drift out of sync. Playing the same file in an external player is perfectly synced, so the desync is introduced by our remux/seek pipeline, not the source file.

Root cause — the two media live on **independently-positioned timelines**:

- **Audio + video** are remuxed by ffmpeg into fragmented MP4 fed to a `SourceBuffer`. The fmp4 muxer normalizes its output to start at PTS 0, so the renderer maps fragments back onto the absolute file timeline with `sb.timestampOffset = keyframeTime` (`use-mse-player.ts:407`, and the initial-open path at `:140`). `keyframeTime` comes from a *separate* ffprobe keyframe scan in main (`findKeyframeBefore`, `streaming/index.ts:274`).
- **Subtitles** are rendered by libass / `SubtitlesOctopus` (`use-subtitles.ts`) strictly against `video.currentTime`, i.e. the *true* absolute file time — they never pass through the remux/offset step.

The bug only manifests **after a skip** because on playback from 0 the offset is 0 (`timestampOffset` is never even set — `use-mse-player.ts:140` guards `keyframeTime > 0`), so both timelines share the origin and stay locked. The first time `keyframeTime > 0` (a forward skip, or a resume-from-middle), any error in the `timestampOffset = keyframeTime` re-mapping becomes a constant audible/visible subtitle offset.

The error: `timestampOffset` is set to the keyframe's **PTS** (from an independent ffprobe scan), but ffmpeg's `-ss <keyframeTime>` + `-c copy` + `-avoid_negative_ts make_zero` (`streaming/index.ts:382,394`) zeroes its output relative to the first **DTS** across streams. With B-frames (DTS < PTS) and audio-copy packet alignment, ffmpeg's real PTS-0 reference ≠ the probed keyframe PTS, so the A/V buffer lands slightly off the absolute timeline while libass keeps drawing subs at the true `currentTime`. External players demux the original file on one shared timeline, so they never hit this.

Note: both open paths (`player.ipc.ts:303`, `:407`) and the seek path (`:486`) feed the **same** probed keyframe value to ffmpeg `-ss` and to the renderer's `timestampOffset`, so they're internally consistent today — the narrow DTS/PTS-after-`make_zero` delta is the only residual error on the copy path.

## Approach & Architecture

**Measurement-first.** The exact magnitude and origin of the delta (B-frame reorder vs. audio-copy start vs. keyframe-scan mismatch) should be confirmed by instrumentation before committing to one correction, then pick whichever the data proves correct. Two candidate robust fixes, both in scope:

1. **Renderer self-correction (codec-agnostic).** After a respawn, once the **first** fragment is appended (before any eviction runs — see plan), read `sb.buffered.start(0)` and compare it to the intended `keyframeTime` we set as `timestampOffset`. The delta `actualStart - keyframeTime` is the exact amount the A/V buffer is off the absolute timeline. Apply that delta as a libass subtitle time offset via the live-read `octopusInstance.timeOffset` field (the lib re-applies `video.currentTime + timeOffset` on every tick — `subtitles-octopus.js:1139/1157`). Do **not** use `setCurrentTime` — it is overwritten on the next video tick. Self-correcting and independent of codec/container quirks.

2. **Main-side exact PTS-0 (true fix at the source).** ffmpeg still seeks to the real keyframe **K** (`-ss K`, unchanged — it must decode valid content from the right place); what changes is the value the renderer assigns to `timestampOffset`. Today that value is also `K`, which silently assumes the keyframe's post-`make_zero` output PTS (`emittedStart`) is exactly 0. With B-frames it isn't, so set `timestampOffset = K − emittedStart` instead, mapping the A/V fragments *exactly* onto the absolute file timeline. (Note the magnitude: this is ~`K` — near the seek time — **not** the bare ~tens-of-ms reorder delay; an implementer setting the offset to a small value has it wrong. Measurement-first pins the exact value and sign.) With A/V genuinely aligned, `video.currentTime` becomes honest and **no** libass subtitle compensation is needed — this is what earns the "fix at the source" name and is the real distinction from option 1 (which leaves A/V off by the delta and compensates subtitle-side). **Crucially, `-ss` and the returned offset are now two different numbers and must be decoupled** — today `player.ipc.ts:486–487` feeds one `keyframeTime` variable both to `spawnFfmpegForSession` (`-ss`) and to the return value (`timestampOffset`); the open paths share `startSeek`/`initialSeek` the same way (`:303`, `:407`). The trade-off: changing `timestampOffset` shifts the content-vs-`currentTime` mapping relative to today, so this option **moves `currentTime`** by the delta and the watch-progress save/resume (reads off `currentTime`) + Syncplay peer-sync audit from finding #4 genuinely applies and must be done. Also more sensitive to ffmpeg version / codec behavior than opt 1 (the delta is *derived* main-side rather than *measured* renderer-side).

The implementer should land instrumentation (option below) first, read the real delta on the copy path and the transcode path, then choose 1 or 2 (or both — a small source-side correction plus a renderer assertion). Whatever lands must keep `video.currentTime` semantics consistent for watch-progress save/resume.

Scope covers **both the stream-copy seek path and the HEVC→H.264 transcode respawn path** (`spawnFfmpegForSession`, `streaming/index.ts:364` — transcode branch adds `-avoid_negative_ts make_zero` too at `:382`), plus resume-from-middle (any `keyframeTime > 0`).

## UI / UX Considerations

No new UI. Behavior change only: subtitles stay in sync with audio/video after a skip. Should be invisible to the user except that the bug is gone.

## Implementation Plan

   - [ ] Add temporary instrumentation to quantify the delta on the copy and transcode paths. ffmpeg's stdout is the binary fmp4 stream piped straight to the renderer (`spawnFfmpegForSession`) and stderr progress lines don't carry the first DTS, so do **not** try to "parse from ffmpeg" — the two feasible measurements are (a) an `ffprobe` of the first emitted fragment, or (b) the renderer's `sb.buffered.start(0)` (which option 1 already reads). Log the probed `keyframeTime` returned from `PLAYER_STREAM_SEEK` (`player.ipc.ts:487`) alongside one of those.
   - [ ] If fix option 2 is chosen: **decouple the two values** in `player.ipc.ts` (and the open paths) — keep `-ss K` at the real keyframe so ffmpeg decodes valid content, but return `K − emittedStart` (keyframe absolute time minus its post-`make_zero` output PTS) as the value the renderer sets as `timestampOffset`. `spawnFfmpegForSession` and the return value must no longer share one variable. This aligns A/V to absolute time — no subtitle compensation, but `currentTime` moves (do the watch-progress/Syncplay audit). Verify across `-c copy` and the transcode branch (both pass `-avoid_negative_ts make_zero`).
   - [ ] No new channel/field — `PLAYER_STREAM_SEEK` already returns `{ generation, keyframeTime }`. Option 2 decouples the two uses of that value: ffmpeg `-ss` keeps the real keyframe **K**, while the returned `keyframeTime` (→ `timestampOffset`) carries the `K − emittedStart` correction. The signature is unchanged, but `spawnFfmpegForSession` and the return value must stop sharing one variable. If option 1's measured delta ever needs to round-trip, extend the return type in `src/shared/types/*.d.ts` (ambient, still no new channel).
   - [ ] If fix option 1 is chosen: in `use-mse-player.ts` `handleUnbufferedSeek`, reset a "measure pending" flag/counter at the start of the respawn, then capture `sb.buffered.start(0) - keyframeTime` on the **first** post-respawn append **before** any eviction runs. `onSourceBufferUpdateEnd` (`:163`) is a single handler shared by initial-open and every respawn and it mutates `buffered.start(0)` via `sb.remove(...)` (lines 176–181) and the `QuotaExceededError` eviction path (`pumpAppendQueue`), so the capture must happen ahead of those. Expose the measured delta.
   - [ ] In `use-subtitles.ts`, store the correction in the composable and apply it via `octopusInstance.timeOffset`. Because `subscribeStreamSubtitles` destroys + re-inits the octopus instance when stream-extracted subs arrive (`:98–99`) — which resets `timeOffset` to 0 — re-apply the stored correction on **every** `initSubtitles`, not just once after a seek.
   - [ ] Apply the same correction to the **initial-open** path when `keyframeTime > 0` (resume-from-middle), `use-mse-player.ts:140` / `PlayerView` `startMseSession` callsites (`:765`, `:861`).
   - [ ] Ensure transcode respawns (`transcodingHevc`) get the correction too.
   - [ ] Remove instrumentation logging once the fix is verified.

## Files to Touch
- **Modify:** `src/renderer/src/composables/use-mse-player.ts` (per-respawn delta measurement before eviction)
- **Modify:** `src/renderer/src/composables/use-subtitles.ts` (store correction, apply `timeOffset` on every init)
- **Modify:** `src/main/streaming/index.ts` (`findKeyframeBefore` / true-start derivation, if option 2)
- **Modify:** `src/main/ipc/player.ipc.ts` (seek handler return, instrumentation)
- **Modify (maybe):** `src/renderer/src/components/views/PlayerView.vue` (wire correction through startMseSession callsites)
- **Documentation:** `docs/player.md` (note the timestampOffset ↔ subtitle-clock alignment behavior)

## Testing Strategy

Automated:
- [ ] Unit test in `use-mse-player.ts`'s test seam (`_internal.handleUnbufferedSeek`, `:580`): drive a respawn with a fixture where the appended fragment's `buffered.start(0)` differs from the requested `keyframeTime`, and where an eviction would fire — assert the captured correction equals the pre-eviction delta (fails on the old behavior where no correction is computed → passes after the fix). This is the regression test that captures the behavior difference.
- [ ] Unit/contract test for the main-side seek return (if option 2): for a B-frame fixture, assert (a) the `-ss` arg passed to `spawnFfmpegForSession` is still the **real keyframe K** (not the corrected value), and (b) the returned `keyframeTime` (→ `timestampOffset`) is the `K − emittedStart` correction — i.e. the two are decoupled. Plus a watch-progress/resume test asserting the `currentTime` shift is accounted for.
- [ ] `use-subtitles.ts` test: given a stored correction, assert `octopusInstance.timeOffset` is set to it on `initSubtitles`, and that a destroy+re-init (stream-subtitle arrival) re-applies it rather than resetting to 0.

Manual:
- [ ] Cached MKV with B-frames (typical anime): play from 0 (confirm still synced), then skip forward several times — subs stay locked to audio. Repeat resume-from-middle.
- [ ] HEVC file via the transcode path — same skip test.
- [ ] Files with and without an audio delay / non-zero start_time.
- [ ] Platforms: Windows + Linux/WSL (the existing respawn stutter notes in `use-mse-player.ts` call out WSL specifically).

## Risks & Edge Cases
- ffmpeg version / codec differences: the DTS-vs-PTS delta may vary by encoder and container; the measurement-first approach mitigates guessing. Option 1 (measure actual buffered start) is the most robust against this.
- Must not break `video.currentTime`-based watch-progress save/resume (`PlayerView` reports position off `currentTime`) and Syncplay peer sync. Option 1 shifts only the subtitle clock (`timeOffset`), so `currentTime` stays honest; option 2 deliberately moves `currentTime` by the delta to align A/V to absolute time, so it **must** audit/adjust the watch-progress save/resume and Syncplay paths (the delta is small, but verify it isn't silently wrong).
- Repeated rapid seeks: the correction must be recomputed per respawn and not accumulate (reset the measure flag on each `handleUnbufferedSeek`).
- Buffer eviction/quota paths (`onSourceBufferUpdateEnd` `sb.remove`, `pumpAppendQueue` `QuotaExceededError`) move `buffered.start(0)` — the measurement must be taken on the first post-respawn append before they run (now an explicit plan step, not just a risk).
- Subtitle re-init: `subscribeStreamSubtitles` destroy+re-init resets `timeOffset` to 0; the stored correction must be re-applied on every `initSubtitles`.

## Out of Scope
- Migrating subtitle rendering from libass-wasm (SubtitlesOctopus) to JASSUB (tracked separately).
- Any non-MKV / CDN-stream subtitle behavior.
- The legacy full-remux fallback path (only the MSE streaming path is in scope).
- General A/V-vs-A/V sync issues unrelated to seeking.

mikkerlo:
Fixes #198

## What

Subtitles drift out of sync with audio/video after seeking (or resuming from the middle) in a cached MKV played through the MSE player. Root cause: ffmpeg's fmp4 muxer emits the seeked keyframe at a small non-zero output PTS (`emittedStart`, the B-frame reorder delay) even with `-avoid_negative_ts make_zero`, so with `sb.timestampOffset = keyframeTime` the A/V content of true file time T lands on the MSE timeline at `T + emittedStart`. libass renders strictly against the honest `video.currentTime`, so it never sees this shift — a constant desync appears the first time `keyframeTime > 0`.

This implements **Option 1** from the issue (renderer-side self-correction): codec-agnostic, keeps `video.currentTime` honest (no watch-progress / Syncplay audit needed), and measures the delta at runtime.

## How

- **`use-mse-player.ts`** — measures `subtitleCorrection = keyframeTime - sb.buffered.start(0)` (= `-emittedStart`) on the **first** post-respawn append (and on the initial-open path when `keyframeTime > 0`), captured **before** any eviction runs (`onSourceBufferUpdateEnd`'s `sb.remove` and `pumpAppendQueue`'s `QuotaExceededError` path both move `buffered.start(0)`). A `measurePending` flag is reset at the start of `handleUnbufferedSeek` and armed once the new run's `timestampOffset` is set. Exposes the value as a reactive `subtitleCorrection` ref.
- **`use-subtitles.ts`** — stores the correction and applies it to the live-read `octopusInstance.timeOffset` field (not `setCurrentTime`, which is overwritten on the next video tick). Re-applies on **every** `initSubtitles`, since `subscribeStreamSubtitles` destroys + re-inits the octopus instance (resetting `timeOffset` to 0) when stream-extracted subs arrive.
- **`PlayerView.vue`** — watches `subtitleCorrection` into `setSubtitleCorrection`. Both the stream-copy and HEVC→H.264 transcode respawn paths feed the same ref.

Only the subtitle clock shifts; `video.currentTime` is untouched.

### Sign convention

The keyframe at absolute time `K` is emitted at output PTS `emittedStart = buffered.start(0) - K`, so the A/V content sits `emittedStart` later than `currentTime`. libass draws at `video.currentTime + timeOffset`, so it lines up when `timeOffset = -emittedStart = K - buffered.start(0)`. That is the stored `subtitleCorrection`, written directly into `octopusInstance.timeOffset`.

## Tests

- `use-mse-player.test.ts`: drives a respawn where the appended fragment's `buffered.start(0)` differs from the requested `keyframeTime` and where an eviction fires — asserts the captured correction equals the **pre-eviction** delta (and that eviction did move `buffered.start(0)` afterwards). Fails on the old behavior (no correction computed).
- `use-subtitles.test.ts`: asserts `octopusInstance.timeOffset` is set to the stored correction on `initSubtitles`, applied live to an existing instance, and re-applied across a destroy + re-init (stream-subtitle arrival) rather than resetting to 0.

`npm run typecheck`, `npm run test` (606 passed), and `npm run format` all pass.

## Docs

`docs/player.md` documents the `timestampOffset` ↔ subtitle-clock (`timeOffset`) alignment behavior.


mikkerlo:
@akvasha — implementation of #198 is up for review. Went with **Option 1** (renderer-side self-correction), since the measurement happens at runtime (no manual delta measurement with real files needed), it's codec-agnostic, and it keeps `video.currentTime` honest — so the watch-progress/Syncplay audit that Option 2 required doesn't apply.

Highlights tying back to your review points on the issue:
- **Pre-eviction capture** — `subtitleCorrection` is measured on the first post-respawn append *before* both eviction sites (`onSourceBufferUpdateEnd`'s `sb.remove` and the `QuotaExceededError` path). The regression test deliberately fires an eviction afterward and asserts the captured value is the pre-eviction `-0.5`, not the post-eviction `-40`.
- **`timeOffset`, re-applied on every init** — written to the live-read `octopusInstance.timeOffset` field (not `setCurrentTime`), and re-applied in `initSubtitles` so the destroy+re-init in `subscribeStreamSubtitles` doesn't drop it.
- **Sign**: `timeOffset = keyframeTime − buffered.start(0) = −emittedStart` (small, ~reorder-delay magnitude); `currentTime` never touched.

Verified locally: `typecheck` ✓, `test` ✓ (606 passed), `format:check` ✓. Patch bump 4.1.16 → 4.1.17, `docs/player.md` updated.

akvasha:
## Review

Faithful implementation of Option 1 from #198 — `use-mse-player` measures `subtitleCorrection = keyframeTime − buffered.start(0)` (= `−emittedStart`) on the first post-respawn append, captured **before** eviction perturbs `buffered.start(0)`, and `use-subtitles` writes it to libass's live-read `octopusInstance.timeOffset` so only the subtitle clock shifts. I traced the arming/measurement ordering (incl. `measurePending = false` at `handleUnbufferedSeek` entry so the buffer-clear's `updateend` can't measure prematurely), the dedup against both eviction paths (`onSourceBufferUpdateEnd`'s `sb.remove` and the `QuotaExceededError` path), the sign derivation, and the re-apply-on-every-`initSubtitles` logic. Core mechanism is correct and the regression tests are solid.

Two findings — neither is a crash; the second is minor.

### 1. Seek-while-paused never renders the correction until playback resumes
`src/renderer/src/composables/use-subtitles.ts` (`applyCorrection` / `setSubtitleCorrection`)

`setSubtitleCorrection` only writes the `timeOffset` field. libass redraws **solely** on video events (`timeupdate`/`seeked`/`playing`/… — `subtitles-octopus.js:1178–1184`). After an unbuffered seek the buffer-ahead gate keeps a paused player paused — `waitForBufferAhead` calls `play()` only when `!wasPaused` (`use-mse-player.ts:297,309`) — and the correction is applied *after* the seek's `seeked` already fired. So a user who scrubs **while paused** (e.g. to inspect a line of dialogue) sees subtitles at the old, uncorrected offset — the exact `emittedStart` desync this PR fixes — until they press play. The unit tests assert the field value but not a redraw, so this slips through.

**Fix:** trigger one redraw after `applyCorrection` (e.g. `octopusInstance.setCurrentTime(video.currentTime + subtitleCorrection)`), at least when the video is paused.

### 2. Untyped `timeOffset` write fails silently on a libass-wasm upgrade
`src/renderer/src/composables/use-subtitles.ts` (`applyCorrection`)

The `(octopusInstance as unknown as { timeOffset: number }).timeOffset = …` cast bypasses the typings, and assigning a plain property never throws, so a future `libass-wasm` bump that renames or stops reading `timeOffset` would silently disable the correction — no error, no failing test (the mock just stores the field). Low severity / maintainability: worth a comment pinning the relied-on vendored behavior, or a test that exercises the real lib rather than a permissive mock.

---

Everything else — measurement gating, sign convention, transcode-path coverage, reset-on-new-session, and `currentTime` honesty — checks out. Finding 1 is the one I'd resolve before merge.

mikkerlo:
Thanks @akvasha — both findings addressed in the latest push (e930c3c).

**1. Paused-scrub redraw (the one to fix before merge).** `applyCorrection` now forces one redraw after writing `timeOffset`: `octopusInstance.setCurrentTime(video.currentTime + subtitleCorrection)` (the lib's render entrypoint, `self.render = self.setCurrentTime`). Since `setSubtitleCorrection` runs *after* the seek's `seeked` fired and the buffer-ahead gate leaves a paused player paused, this makes a paused scrub reflect the shift immediately instead of waiting for the next `play()`. The next live tick recomputes `currentTime + timeOffset` to the same value, so it's idempotent. Added a regression test asserting `setCurrentTime` is called with the corrected time (`99.5` for `currentTime=100`, `correction=-0.5`) — it fails on the field-only version that slipped this through.

**2. Silent failure on libass-wasm upgrade.** Documented the pinned vendored behavior in `applyCorrection` (the `timeOffset` live-read at `subtitles-octopus.js:1009/1139/1157` + that a plain property write never throws), and pointed at the new redraw test as the canary — a rename/drop of the field would surface as a missing `setCurrentTime`/offset shift there rather than passing silently.

Local gate re-run: `typecheck` ✓, `format:check` ✓, `test` ✓ (607 passed).

akvasha:
## Re-review (post e930c3c) — LGTM ✅

Both findings resolved cleanly.

**1. Paused-scrub redraw.** `applyCorrection` now forces one redraw via `setCurrentTime(v.currentTime + subtitleCorrection)` after writing `timeOffset` — so a paused scrub reflects the shift immediately instead of waiting for the next `play()`. I checked the new call paths for regressions: it's null-guarded on the video element, idempotent against the next live tick (recomputes `currentTime + timeOffset` to the same value), and safe when invoked right after construction in `initSubtitles` or mid-playback. The regression test asserting `setCurrentTime(99.5)` (currentTime 100, correction −0.5) fails on the field-only version that slipped this through — good canary.

**2. Silent libass-wasm-upgrade failure.** Pinned-behavior comment in `applyCorrection` + the redraw test as the canary is the right call for an untyped vendor field; a renamed/dropped `timeOffset` or `setCurrentTime` now surfaces as a missing shift in that test rather than passing silently.

Core mechanism (pre-eviction measurement gating, sign convention, transcode-path coverage, reset-on-new-session, `currentTime` honesty) was already correct and is unchanged. No remaining blockers from me — ready to merge. 🚀

mikkerlo:
@akvasha thanks for the LGTM! 🙏 One formality: it came through as a comment, so the branch-protection gate still reads `REVIEW_REQUIRED` and the PR is `BLOCKED`. Could you submit a formal **Approve** review (Files changed → Review changes → Approve) so it clears? Then I'll squash-merge normally — no admin override needed.

akvasha (approved):
Approving — implementation faithfully follows the agreed Option 1 design from #198, the core mechanism (pre-eviction measurement gating, sign convention, transcode-path coverage, reset-on-new-session, `currentTime` honesty) is correct, and both review findings (paused-scrub redraw + untyped-field canary) are resolved in e930c3c with matching regression tests. No remaining blockers. 🚀

